PR #3, opened 2018-06-05
GAWB-3612 Oauth with Fence

gpolumbo-broad:

In all cases:


main.py
```diff
@@ -51,13 +68,16 @@ def __init__(self):
...
     @endpoints.method(
         OAUTH_CODE_RESOURCE,
-        LinkInfoResponse,
+        AccessTokenResponse,
         path='fence/oauthcode',
         http_method='POST',
         name='fence/oauthcode')
     def oauthcode(self, request):
         user_info = self.auth.require_user_info(self.request_state)
-        return LinkInfoResponse(expires=datetime.now(), username=request.oauthcode)
```
dvoet: I think this should return LinkInfoResponse. Do we get any of this information from fence?

dvoet:
I would like to see all the fence functionality in a different file than main with the actual calls to fence written in such a way as they can be mocked.

config.ini
```diff
@@ -0,0 +1,6 @@
+[fence]
+CLIENT_ID=REPLACE_ME
+CLIENT_SECRET=REPLACE_ME
+AUTHZ_URI=https://qa.dcf.planx-pla.net/user/oauth2/authorize
+REDIRECT_URI=https://local.broadinstitute.org/fence-callback
+TOKEN_URL=https://qa.dcf.planx-pla.net/user/oauth2/token
```
dvoet: why an ini file instead of environment variables in app.yml?

main.py
```diff
@@ -43,21 +48,32 @@ class StatusResponse(messages.Message):
...
+        token_response = self.oauth_adapter.exchange_authz_code(request.oauthcode)
+        TokenStore.save(user_info.email, token_response)
```
dvoet: we don't want to use the email as a key, use user_info.id

bond_token.py
```diff
@@ -0,0 +1,38 @@
+from google.appengine.ext import ndb
+import json
+
+
+class BondToken(ndb.Model):
+    token_dict_str = ndb.TextProperty()
+
+    REQUIRED_KEYS = ["access_token", "refresh_token", "token_type"]
```
dvoet: We don't need to store access_token, it is only good for an hour. 

What is token_type and why do we need to store it?

I think we should also store the date it was created and the user name within the other system.

bond_token.py
```diff
@@ -0,0 +1,38 @@
+from google.appengine.ext import ndb
+import json
+
+
+class BondToken(ndb.Model):
```
dvoet: This is an oauth refresh token. BondToken does not make much sense to me.

config.ini
gpolumbo-broad: This was done for testing.  Especially in intellij or I think even using nose, we can always read from the .ini, whereas app.yml isn't always loaded.

PR #6, opened 2018-06-13
Dockerize bond

MatthewBemis:
This will autobuild and is the first step towards having Bond usable in FiaB.

In all cases:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
       - checkout
       - run: # install and activate virtual environment with pip
           command: |
-	    rm Dockerfile
+            rm Dockerfile
             python2.7 -m virtualenv virtualenv
             . virtualenv/bin/activate
             pip install -t lib/ -r requirements.txt
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,7 @@ jobs:
       - checkout
       - run: # install and activate virtual environment with pip
           command: |
+	    rm Dockerfile
             python2.7 -m virtualenv virtualenv
             . virtualenv/bin/activate
             pip install -t lib/ -r requirements.txt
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:2.7.15
```
MatthewBemis: would like to trim this down at some point because this image is 1.61gb

README.md
```diff
@@ -4,3 +4,6 @@ Account linking service
 # Running Tests
 * `pip install nose nosegae nose-exclude`
 * ```nosetests --with-gae --gae-lib-root=`gcloud info --format="value(installation.sdk_root)"`/platform/google_appengine --exclude-dir=lib```
+
+# Running locally in Docker
+* `docker run -p=8080:8080 -p=8000:8000 quay.io/databiosphere/bond:latest`
```
MatthewBemis: 8000 is exposed and published...but for some reason it serves up an empty admin console

our APIs at 8080 do still work, though. so for now it's workable but i'd like to get this figured out

.circleci/config.yml
```diff
@@ -15,6 +15,7 @@ jobs:
       - checkout
       - run: # install and activate virtual environment with pip
           command: |
+            rm Dockerfile
```
MatthewBemis: there is this nice "feature" that detects a dockerfile and makes the assumption that you're trying to use a custom runtime environment in GAE. that is not actually the case here.

for building the docker image, the `.dockerignore` is sufficient for getting past this because the Dockerfile isn't used in the image itself. for circle, we need to actually remove the Dockerfile. fortunately it has no use in either of these cases.

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# Prepare the image
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt-get update && apt-get install -y -qq --no-install-recommends wget unzip python python-pip python-dev build-essential openssh-client python-openssl && apt-get clean
```
gpolumbo-broad: I'd feel a lot better if we specified python 2.7 here

PR #9, opened 2018-06-19
tag develop when circle tests pass

MatthewBemis:

In all cases:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,4 +49,4 @@ workflows:
             - test
           filters:
             branches:
-              only: mb-tag-on-success
+              only: develop
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,4 +49,4 @@ workflows:
             - test
           filters:
             branches:
-              only: develop
+              only: mb-tag-on-success
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,10 +7,10 @@ jobs:
   test:
     docker:
       - image: quay.io/databiosphere/bond-tests:latest
-      
+
     environment:
       - PYTHONPATH: /usr/local/google_appengine
-      
+
     steps:
       - checkout
       - run: # install and activate virtual environment with pip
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -21,9 +21,33 @@ jobs:
...
-          
+
+  tag:
+    docker:
+      - image: alpine/git
+
+    steps:
+      - add_ssh_keys:
+          fingerprints:
+            - "ab:2d:12:de:d1:ac:36:0b:6a:02:92:3f:2c:49:98:c9"
```
jacmrob: is this safe to be putting into source?  (i'm no ssh expert :( )

.circleci/config.yml
```diff
@@ -21,9 +21,33 @@ jobs:
...
+      - checkout
+      - run:
+          when: on_success
+          command: |
+            echo 'export TAG_NAME=$(date +%s)' >> $BASH_ENV
```
jacmrob: does BASH_ENV come from the circle env vars, or are you expecting to pass it in somehow?

.circleci/config.yml
```diff
@@ -21,9 +21,33 @@ jobs:
...
+            source $BASH_ENV
+            git --version
```
jacmrob: necessary?

.circleci/config.yml
```diff
@@ -21,9 +21,33 @@ jobs:
...
+            git tag dev_tests_passed_$TAG_NAME
```
dvoet: put the dev_tests_passed_ prefix in the export statement above

.circleci/config.yml
MatthewBemis: im just learning this now but the fingerprint is a hash of the public key so it's okay to publicize

.circleci/config.yml
MatthewBemis: i think it's being created within the job. was following these instructions: https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-shell-command

PR #14, opened 2018-06-21
Deploy scripts

MatthewBemis:

In all cases:


dvoet:
there is also the config.ini which should be templatized

app.yaml.ctmpl
```diff
@@ -28,8 +30,13 @@ libraries:
 env_variables:
   # The following values are to be replaced by information from the output of
   # 'gcloud endpoints services deploy swagger.json' command.
-  ENDPOINTS_SERVICE_NAME: YOUR-PROJECT-ID.appspot.com
-  ENDPOINTS_SERVICE_VERSION: 2016-08-01r0
+  ENDPOINTS_SERVICE_NAME: {{$googleProject}}.appspot.com
```
jacmrob: you need to define this in the top of the tmpl file

app.yaml.ctmpl
```diff
@@ -28,8 +30,13 @@ libraries:
...
+  ENDPOINTS_SERVICE_VERSION: {{$serviceVersion}}
```
jacmrob: same as above but with an env var called `SERVICE_VERSION` or something

manifest.rb
```diff
@@ -0,0 +1,2 @@
+render "app.yaml.ctmpl"
```
jacmrob: so using manifest.rb is a relic of using the configure.rb script, which has to be in the repo for this to work.  I think for indie deploys we may want to go a different route, since configure depends on the configs looking a certain way and having certain environment variables.  instead, we should make all rendering happen within the deploy.sh script and use the script in dsde-toolbox, which allows for any "environment" as long as you pass it in during the docker run.

so to render app.yaml.ctmpl you would put this in `deploy.sh`:
`docker run -v $PWD:/app -e GOOGLE_PROJ=$proj -e SERVICE_VERSION=$version -e INPUT_PATH=/app -e OUT_PATH=/app/output render-templates.sh`
this would render out to a directory called "output" in the root of the repo, and would render all ctmpl files in the directory we mounted as a volume, which is currently the root (`$PWD`).  to clean things up though we would probably want to make a configs directory in the root of the repo itself and then pass that as a volume, rather than the whole repo.

app.yaml.ctmpl
jacmrob: add `{{with $googleProj := env "GOOGLE_PROJ"}}` to the top and another `{{end}}` clause to the end.  this expects an env var called `GOOGLE_PROJ` to exist when rendering; see my below comment about rendering for how to pass that through

manifest.rb
jacmrob: https://github.com/broadinstitute/dsde-toolbox/blob/master/render-templates.sh if you want to know how that script works

deploy.sh
```diff
@@ -0,0 +1,22 @@
...
+SERVICE_VERSION=`gcloud endpoints services describe $GOOGLE_PROJECT.appspot.com --format=json | jq .serviceConfig.id`
+
+#render app.yaml and config.ini for environment with SERVICE_VERSION and GOOGLE_PROJECT
+docker run -v $PWD:/app -e GOOGLE_PROJ=$GOOGLE_PROJECT -e SERVICE_VERSION=$SERVICE_VERSION -e INPUT_PATH=/app -e OUT_PATH=/app/output render-templates.sh
+
+#install dependencies to lib folder
+pip install -t lib -r requirements.txt
+
+#deploy the app to the specified project
+gcloud -q app deploy app.yaml --project=$GOOGLE_PROJECT
```
jacmrob: this will need to be `gcloud -q app deploy output/app.yaml --project=$GOOGLE_PROJECT`, if we are rendering configs to `output` directory

jacmrob: I can change as I need to test git push triggers on this repo in jenkins

deploy.sh
```diff
@@ -0,0 +1,33 @@
...
+GOOGLE_PROJECT=$3
+
+#jenkins needs access to the projects so it can deploy
+# command TBD (that's either already the case or we use some credential here to make it so)
+
+#render the endpoints json
+python lib/endpoints/endpointscfg.py get_openapi_spec main.BondApi --hostname $GOOGLE_PROJECT.appspot.com
+
+#deploy google endpoints
+gcloud -q endpoints services deploy linkv1openapi.json
```
jacmrob: in order to make all gcloud commands work you'll have to authenticate with a service account, which you should grab from vault.  see how we do it during fiab populate, for example - https://github.com/broadinstitute/firecloud-develop/blob/dev/run-context/fiab/scripts/fiab-populate/sam_google_extensions_security.sh#L25

then you'll want to run `gcloud auth activate-service-account <path-to-svc-acct>`

deploy.sh
```diff
@@ -0,0 +1,33 @@
+VAULT_TOKEN=$1
+ENVIRONMENT=$2
+
+# TODO: this needs to be derived from the env somehow; the only two vars this script can take are VAULT_TOKEN and ENVIRONMENT
```
jacmrob: FYI

deploy.sh
```diff
@@ -0,0 +1,32 @@
+VAULT_TOKEN=$1
+ENVIRONMENT=$2
+GOOGLE_PROJECT=broad-bond-$ENVIRONMENT
+
+docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN -e VAULT_ADDR=${VAULT_ADDR} broadinstitute/dsde-toolbox vault read --format=json "secret/dsde/bond/dev/deploy-account.json" | jq .data > deploy_account.json
```
jacmrob: probably want to make the "dev" in the path conditional

app.yaml.ctmpl
```diff
@@ -0,0 +1,42 @@
...
+  version: 2.6
+- name: ssl
+  version: 2.7.11
+
+# [START env_vars]
+env_variables:
+  # The following values are to be replaced by information from the output of
+  # 'gcloud endpoints services deploy swagger.json' command.
+  ENDPOINTS_SERVICE_NAME: {{$googleProject}}.appspot.com
+  ENDPOINTS_SERVICE_VERSION: 2018-06-28r6
```
MatthewBemis: this is a temp solution, will parse this out of `gcloud endpoints services deploy` output until google fixes the `describe` command

dvoet: I would templatize this now and feed in the temp value

PR #15, opened 2018-06-22
status

dvoet:

In all cases:

### DIFF
--- a/status.py
+++ b/status.py
@@ -1,0 +1,62 @@
+from google.appengine.api import memcache
+
+class Status:
+    def __init__(self, fence_api, sam_api):
+        self.fence_api = fence_api
+        self.sam_api = sam_api
+
+    def get(self):
+        """
+        gets status of all sub systems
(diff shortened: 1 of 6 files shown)


status.py
```diff
@@ -0,0 +1,62 @@
...
+                    Subsystems.fence: {"ok": fence_ok, "message": fence_message},
+                    Subsystems.sam: {"ok": sam_ok, "message": sam_message}
+                }
+                self._cache_status(status)
+            return status
+        except Exception as e:
+            # any exception at this point is memcache
+            return {
+                Subsystems.memcache: {"ok": False, "message": e.message},
+                Subsystems.datastore: {"ok": False, "message": "unknown"},
```
MatthewBemis: memcache being non-functional doesn't necessarily mean that the 3 other subsystems are down. we could still return the status for them. it'd be best to know the complete story during any outage

dvoet: I am going to leave it as is. If memcache is down and we query other systems we will hammer them.

MatthewBemis: :+1:

PR #34, opened 2018-07-25
support multiple oauth providers

dvoet:
Introduced the concept of providers. Each section in config.ini (except sam) represents a provider. Bond and FenceTokenVendingMachine take the provider name in their init methods and use that to segregate data in datastore and memcache. Status also checks each provider and I had to change the output format of status api because protorpc can't do dictionaries that well :(

In all cases:


.dockerignore
```diff
@@ -0,0 +1 @@
+/lib
```
dvoet: this prevents the local lib dir from getting into the docker

docker/Dockerfile
```diff
@@ -14,9 +14,11 @@ RUN mkdir /.ssh
 ENV PATH /google-cloud-sdk/bin:$PATH
 VOLUME ["/.config"]
 
-ADD . app/
+RUN mkdir /app
 WORKDIR app/
+ADD requirements.txt .
 RUN pip install -t lib/ -r requirements.txt
+ADD . .
```
dvoet: reordered so that pip install doesn't have to happen every time when code changes and we rebuild

main.py
```diff
@@ -42,123 +46,153 @@ class ServiceAccountAccessTokenResponse(messages.Message):
...
-        self.fence_tvm = FenceTokenVendingMachine(fence_api, sam_api, oauth_adapter)
-        self.bond = Bond(oauth_adapter, fence_api, sam_api, self.fence_tvm)
+    @endpoints.method(
+        PROVIDER_RESOURCE,
+        ListProvidersResponse,
+        path='/',
+        http_method='GET',
+        name='listProviders')
+    def providers(self, request):
+        return ListProvidersResponse(providers=self.providers.keys())
```
dvoet: new api so we can list configured providers

PR #49, opened 2018-08-09
tag green dev as latest

MatthewBemis:

In all cases:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,6 +36,8 @@ jobs:
             source $BASH_ENV
             git tag $TAG_NAME
             git push origin $TAG_NAME
+            git tag dev_tests_passed_latest
+            git push origin dev_tests_passed_latest
 
 workflows:
   version: 2


.circleci/config.yml
```diff
@@ -36,6 +36,8 @@ jobs:
             source $BASH_ENV
             git tag $TAG_NAME
             git push origin $TAG_NAME
+            git tag dev_tests_passed_latest
```
gpolumbo-broad: If this has a static name like this, I think you'll need to delete the tag first http://www.manikrathee.com/how-to-delete-a-tag-in-git.html, push the deletion, and then recreate the tag?  I think?

MatthewBemis: good catch, will re-work this

gpolumbo-broad: So, I was going a bit more elaborate/overboard maybe with what I was thinking.  That's fine though.  I think it's important that we end up with official release tags though.  So you might want to update the README too to add a step to the release process to create a semanticly versioned tag when starting the release process.  This is to try to avoid a scenario where you're doing a release and trying to merge `dev_tests_passed_latest` into staging or something, and I push a commit to dev and circle moves the `dev_tests_passed_latest` tag on you in the middle of your release. 

Another thought was that we put some sort of `.version` file into the root of the repo.  That way, the engineer making the changes in responsible for bumping the version number appropriately.  Then, instead of a statically named tag being created by circle, we do something like, `git tag $(cat .version)-$BUILD_NUMBER`.

PR #70, opened 2018-12-05
Circle builds docker image and runs tests

gpolumbo-broad:
This PR does a couple of things:

1. CircleCI now does the `docker build` step instead of letting quay.io do this
1. This was done so that we can run some functional tests pointing Bond at the [mock-provider](https://github.com/broadinstitute/mock-provider)

**Something we'll need to check**
I am not sure how the `latest` tag used to be created by quay.io when it was responsible for building the docker images.  We may need to add a step to push this tag when building the dev branch.

In all cases:

### DIFF
--- a/tests/integration/oauth_adapter_test.py
+++ b/tests/integration/oauth_adapter_test.py
@@ -54,7 +54,7 @@ class OauthAdapterTestCase(unittest.TestCase):
             print("Please go to %s to authorize access: %s" % (provider, authz_url))
             print("Please copy/paste the \"code\" parameter from the resulting URL: ")
             sys.stdout.flush()
-            auth_code = sys.stdin.readline().strip()
+            # auth_code = sys.stdin.readline().strip()
             authz_responses[provider] = oauth_adapter.exchange_authz_code(auth_code, redirect_uri)
         local_tb.deactivate()
         return authz_responses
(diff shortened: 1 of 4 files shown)


tests/integration/oauth_adapter_test.py
```diff
@@ -54,7 +54,8 @@ def authorize_with_providers(cls, oauth_adapters):
             print("Please go to %s to authorize access: %s" % (provider, authz_url))
             print("Please copy/paste the \"code\" parameter from the resulting URL: ")
             sys.stdout.flush()
-            auth_code = sys.stdin.readline().strip()
+            # auth_code = sys.stdin.readline().strip()
+            auth_code = "X"
```
dvoet, line 58: ?

dvoet (approved):

marctalbott (approved):

tests/integration/oauth_adapter_test.py
gpolumbo-broad: So, before we had the mock-provider, the tests would wait for me to click the authz link, get an access code, and then paste it in here.  This was handy for confirming that Bond worked with  the Fences.  With the mock-provider in place, the access code is irrelevant, the mock-provider always authorizes.  

I left the code commented out so that we could still uses if we want to test against a real provider for some reason.

PR #88, opened 2020-01-16
[CA-592] Migrate from google endpoints to flask in python2

yonghaoy:
Migrate endpoint to flask in python2, including exceptions, request format and routes. 
I am not chaning the response and it is still using protoprc.message. Will change it if we have to, and it depends on Andrew's json test, to see if there is any difference in response. 

In all cases:


routes.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+routes = Blueprint('bond', __name__, '/')
+
+bond_providers = {provider_name: create_provider(provider_name)
+                  for provider_name in config.sections() if provider_name != 'sam'}
+auth = authentication.Authentication(authentication.default_config())
+
+
+@routes.route('/api/link/v1/providers', methods=["GET"])
```
marctalbott: You asked me about this already, but it'd be nice to have some kind of route hierarchy rather than having a flat set of routes. It doesn't look like Flask really supports that, but it might be nice to set `'/api/link/v1'` equal to `api_routes_base` or something like that so that each of these routes would be a bit more grouped at least.  
```suggestion
api_routes_base = '/api/link/v1'
@routes.route(api_routes_base + '/providers', methods=["GET"])
```

requirements.txt
```diff
@@ -6,4 +6,6 @@ requests-toolbelt==0.8.0
 pyjwt==1.6.4
 mock==2.0.0
 dpath==1.4.2
-google-auth==1.6.1
\ No newline at end of file
+google-auth==1.6.1
+Flask==1.1.1
+Werkzeug==0.16.0
```
marctalbott: At what point can we remove the `google-endpoints-api-management` requirement? It seems like it is also only compatible with python 2, so we may need to find alternatives for `oauthclient` and `urllib`

routes.py
```diff
@@ -0,0 +1,194 @@
...
+
+@routes.route('/api/status/v1/status', methods=["GET"])
+def get_status():
+    sam_base_url = config.get('sam', 'BASE_URL')
+
+    providers = {provider_name: FenceApi(config.get(provider_name, 'FENCE_BASE_URL'))
+                 for provider_name in config.sections() if provider_name != 'sam'}
+
+    sam_api = SamApi(sam_base_url)
+    status_service = Status(sam_api, providers)
```
marctalbott, line 205: Is there a way to separate out this startup stuff from the actual status query? Seems like unnecessary (although admittedly pretty trivial) work for each status call

requirements.txt
yonghaoy: I think we can remove api-management if we add oauthclient, urllib into our requirement.txt?

dvoet (approved):

app.yaml.ctmpl
```diff
@@ -19,8 +19,8 @@ skip_files:
 #[END_EXCLUDE]
 
 handlers:
-- url: /api/.*
-  script: main.api
+  - url: /.*
```
andy7i, line 22: just curious - what was this change for?

tests/test_runner.py
```diff
@@ -33,6 +33,11 @@ def fixup_paths(path):
 
 
 def main(sdk_path, test_path, test_pattern):
+    os.environ["current_version_id"] = "testbed.version"
```
andy7i, line 36: should this be in caps?

routes.py
```diff
@@ -0,0 +1,217 @@
...
+
+class Parser(FlaskParser):
+    DEFAULT_VALIDATION_STATUS = 400
+
+
+parser = Parser()
+use_args = parser.use_args
+
+
+class JsonField(messages.StringField):
```
andy7i, line 31: Probably outside the scope of this ticket (since it's existing code), but with all these classes here, I wonder if it would be helpful to make it clear which are public/protected/private attributes so that it's a bit clearer how to / not to use them. Then again, routes is only imported by main.py, so ¯\_(ツ)_/¯

routes.py
```diff
@@ -0,0 +1,217 @@
...
+
+class BondProvider:
+    def __init__(self, fence_tvm, bond):
+        self.fence_tvm = fence_tvm
+        self.bond = bond
+
+
+routes = Blueprint('bond', __name__, '/')
+
+bond_providers = {provider_name: create_provider(provider_name)
```
andy7i, line 122: `bond_providers` and `auth` used to be in a class structure (`self.providers` and `self.auth`). just wondering why we pulled them out

app.yaml.ctmpl
andy7i: maybe answering my own question:

https://github.com/DataBiosphere/bond/pull/88/files#diff-d4aed59f80ec0ed138cdae96389f77c7R126
the routes base url is:
`api_routes_base = '/api/link/v1'`

routes.py
```diff
@@ -0,0 +1,217 @@
...
+
+api_routes_base = '/api/link/v1'
+
+
+@routes.route(api_routes_base + '/providers', methods=["GET"])
+def list_providers():
+    return protojson.encode_message(ListProvidersResponse(providers=list(bond_providers.keys())))
+
+
+@routes.route('/api/link/v1/<provider>/oauthcode', methods=["POST"])
```
andy7i, line 134: could use `api_routes_base` here

routes.py
```diff
@@ -0,0 +1,217 @@
...
+    authz_url = _get_provider(provider).bond.build_authz_url(args['scopes'], args['redirect_uri'], args['state'])
+    return protojson.encode_message((AuthorizationUrlResponse(url=authz_url)))
+
+
+@routes.route('/api/status/v1/status', methods=["GET"])
+def get_status():
+    sam_base_url = config.get('sam', 'BASE_URL')
+
+    providers = {provider_name: FenceApi(config.get(provider_name, 'FENCE_BASE_URL'))
+                 for provider_name in config.sections() if provider_name != 'sam'}
```
andy7i, line 202: this looks pretty similar to:

```
bond_providers = {provider_name: create_provider(provider_name)
                  for provider_name in config.sections() if provider_name != 'sam'}
```

but I don't have an idea of how to optimize that and may be out of scope for this ticket

routes.py
```diff
@@ -0,0 +1,217 @@
...
+def service_account_accesstoken(args, provider):
+    user_info = auth.require_user_info(request)
+    return protojson.encode_message(ServiceAccountAccessTokenResponse(token=_get_provider(provider).fence_tvm.get_service_account_access_token(user_info, args['scopes'])))
+
+
+@routes.route(api_routes_base + '/<provider>/authorization-url', methods=["GET"])
+@use_args({"scopes": fields.Str(repeated=True, missing=None),
+           "redirect_uri": fields.Str(required=True),
+           "state": fields.Str(missing=None)},
+          locations=("querystring",))
```
andy7i, line 191: unary tuples are weird -- thanks for the heads up!

andy7i (approved):
Overall, looks good. Mostly just comments/questions. I like how the flask api definitions are easier to read. I double checked to make sure all uses of `raise endpoints` were updated.

I ran `tests/test_runner.py` on this branch and had some issues getting tests to pass. Maybe I'm missing a step somewhere. The local bond logs say: 

> ImportError: <module 'main' from '/Users/ajang/Firecloud/bond/main.pyc'> has no attribute api

If we can get those tests to pass, 👍

app.yaml.ctmpl
marctalbott: I believe it is just to change how Google App Engine routes requests to Bond

routes.py
marctalbott: This is what I was complaining to Greg about this morning. I don't like that this stuff is just floating on its own now, but it works and moving it into a class would likely cause some headaches from a dependency/import standpoint so I'm inclined to just leave it for now

routes.py
marctalbott: Is that code from somewhere else? I'm not really sure what we'd be optimizing here, and I believe that this stuff was transferred over 1-1 from the old `main.py`

routes.py
marctalbott: Yeah, I was thinking that it'd be nice to pull all of them out of `routes.py` and in to a different file (`responses.py` or something like that) and then import them to `routes.py`, but I think it is probably outside of the scope of this ticket since this was essentially just copied over from `main.py`

tests/test_runner.py
marctalbott: I'm actually going to remove this since I believe it was leftover from some manual testing. Unit tests pass without it

marctalbott:
Closing this because all of the commits in this PR were squashed and included in the new feature branch (`python3-feature`)

PR #90, opened 2020-01-21
Fix automatic docker integration test

wchamber:
Revert 65e3a45e6 which allows the test to pass manually in order to run in integration mode by default. Add some comments about what's going on.

In all cases:

### DIFF
--- a/tests/integration/oauth_adapter_test.py
+++ b/tests/integration/oauth_adapter_test.py
@@ -49,17 +49,21 @@ class OauthAdapterTestCase(unittest.TestCase):
         redirect_uri = "http://local.broadinstitute.org/#fence-callback"
         state = "abc123"
         authz_responses = {}
+        # Whether this test is being run manually or for by automated integration.
+        # The integration run uses a fake provider. The manual run allows testing against real providers with real
+        # authentication. There's not a good way to run integration tests against real providers, so we have this
+        # flag to manually change the tests.
+        manual_run = False
         for provider, oauth_adapter in oauth_adapters.iteritems():
             authz_url = oauth_adapter.build_authz_url(scopes,
                                                       redirect_uri,
                                                       state,
-                                                      extra_authz_url_params={"idp": "google"})
+                                                      extra_authz_url_params={"idp": "google"} if manual_run else {"foo": "bar"})
             print("Please go to %s to authorize access: %s" % (provider, authz_url))
             print("YOU WILL BE REDIRECTED TO %s WHICH WILL PROBABLY UNREACHABLE -- THIS IS EXPECTED!" % redirect_uri)
             print("Please copy/paste the \"code\" parameter from the resulting URL: ")
             sys.stdout.flush()
-            auth_code = sys.stdin.readline().strip()
-            # auth_code = "X"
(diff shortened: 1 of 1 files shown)


tests/integration/oauth_adapter_test.py
```diff
@@ -49,17 +49,21 @@ def authorize_with_providers(cls, oauth_adapters):
...
+        # The integration run uses a fake provider. The manual run allows testing against real providers with real
+        # authentication. There's not a good way to run integration tests against real providers, so we have this
+        # flag to manually change the tests.
+        manual_run = False
         for provider, oauth_adapter in oauth_adapters.iteritems():
             authz_url = oauth_adapter.build_authz_url(scopes,
                                                       redirect_uri,
                                                       state,
-                                                      extra_authz_url_params={"idp": "google"})
+                                                      extra_authz_url_params={"idp": "google"} if manual_run else {"foo": "bar"})
```
gpolumbo-broad: I think we can just leave this as is without the `if` statement all the time

gpolumbo-broad (changes requested):

marctalbott (approved):

tests/integration/oauth_adapter_test.py
```diff
@@ -49,17 +49,21 @@ def authorize_with_providers(cls, oauth_adapters):
...
             print("Please go to %s to authorize access: %s" % (provider, authz_url))
             print("YOU WILL BE REDIRECTED TO %s WHICH WILL PROBABLY UNREACHABLE -- THIS IS EXPECTED!" % redirect_uri)
             print("Please copy/paste the \"code\" parameter from the resulting URL: ")
             sys.stdout.flush()
-            auth_code = sys.stdin.readline().strip()
-            # auth_code = "X"
+            auth_code = sys.stdin.readline().strip() if manual_run else "X"
```
andy7i, line 66: python ternary syntax takes some getting used to :\

Is this line needed just so that in an automated run, it doesn't wait for stdin? Because otherwise won't it fail with an auth code of "X"? / Is this test ever run in automated tests?

andy7i (approved):
1 question on whether we need to handle automated runs currently (afaik, we don't run it in an automated fashion?). Otherwise 👍

tests/integration/oauth_adapter_test.py
wchamber: It is needed so that the automated run will pass, otherwise we get an error, like this branch has now: https://circleci.com/gh/DataBiosphere/bond/593?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

So this is being run by the automated testing in by means of the circleci functional_test.sh call in the 'build-docker-image' flow

tests/integration/oauth_adapter_test.py
wchamber: done

tests/integration/oauth_adapter_test.py
andy7i: makes sense -- thanks!

dvoet (approved):

gpolumbo-broad (approved):

PR #91, opened 2020-01-21
CA-648 Abstract out Fence JWT Token Storage & Locking

wchamber:
Pull out the FenceTokenVendingMachine's logic for interacting with Datastore into a separate class, FenceTokenStorage.

This will allow us to inject a stub for unit testing when we cannot use real Datastore or an emulator. We can no longer rely on testbed's datastore stub as it is not available in python 3 app engine.

In all cases:

### DIFF
--- a/fence_token_storage.py
+++ b/fence_token_storage.py
@@ -3,26 +3,51 @@ import time
 from google.appengine.ext import ndb
 from google.appengine.api.datastore_errors import TransactionFailedError
 
-# How long to use a created value before it is considered expired.
-_VALUE_LIFETIME = datetime.timedelta(days=5)
+# How long to keep a fence service account key before expiring it.
+_FSA_KEY_LIFETIME = datetime.timedelta(days=5)
 
 
-class ServiceAccountNotUpdatedException(Exception):
-    pass
+def create_fence_service_account_key(provider_name, user_id):
+    """Creates a datastore key to associate a (user, provier) to the credentials of a service account fetched from a Fence.
+    :return Returns an ndb Key for the fence service account, i.e. a "fsa_key"
+    """
+    return ndb.Key("User", user_id, FenceServiceAccount, provider_name)
 
 
 class FenceServiceAccount(ndb.Model):
+    """Datastore model for storing fence service account json key and metadata for expiration and generation."""
+    # The string json key token for the fence service account.
     key_json = ndb.TextProperty()
+    # The datetime when to expire the service account key. Only set when the key_json is not None.
     expires_at = ndb.DateTimeProperty()
+    # The datetime of when the current lock on updating this model expires. The model should only be updated by
+    # a service that holds a lock. Only set when the key_json is not set or has expired.
     update_lock_timeout = ndb.DateTimeProperty()
 
 
-class DatastoreLockedStorage:
-    """TODO write me"""
+class ServiceAccountNotUpdatedException(Exception):
+    """An exception for when a fence service account was not updated after some service had grabbed the lock to update it."""
+    pass
+
+
+class FenceTokenStorage:
+    """
+    Stores service account tokens retrieved from fences and manages distributed concurrent updates so that only one access
+    per (user, provider) occurs at a time.
+
+    This is important to prevent many simultaneous requests from overloading a fence when they are all seeking the same
+    service account credentials. Storage of the fence credentials (with an expiration time) allows the retrieved
+    credentials to be shared.
+
+    N.B. "fsa_key" refers to a Datastore key of the (user, provider_name) which is used to look up the fence service
+    account credentials, i.e. "key_json."
+
+    This is abstracted as its own class so that we can stub out the Datastore dependency in tests.
+    """
 
     def delete(self, fsa_key):
         """
-        Delete the stored value for the key.
+        Delete the stored fence service account info for the fsa_key.
         ":return returns the stored value for the key or None if it did not exist.
         """
         fence_service_account = fsa_key.get()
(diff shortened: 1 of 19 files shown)


fence_token_storage.py
```diff
@@ -0,0 +1,157 @@
...
+    def _acquire_lock(self, fsa_key):
+        """
+        :param fsa_key:
+        :return: True if the lock was acquired, False otherwise
+        """
+        try:
+            return self._lock_fence_service_account(fsa_key)
+        # We expect a transaction failure when someone else acquires the lock instead of us. That's fine, it's just
+        # a different way we could fail to acquire the lock.
+        except TransactionFailedError:
```
wchamber: I scoped down this exception to a transaction failure, removing the old TODO

fence_token_storage.py
```diff
@@ -0,0 +1,156 @@
...
+            key_json = fence_fetch_fn(prepped_key)
+            fence_service_account = FenceServiceAccount(key_json=key_json,
+                                                        expires_at=datetime.datetime.now() + _FSA_KEY_LIFETIME,
+                                                        update_lock_timeout=None,
+                                                        key=fsa_key)
+            fence_service_account.put()
+
+        else:
+            fence_service_account = self._wait_for_update(fsa_key)
+            if not fence_service_account.expires_at or fence_service_account.expires_at < datetime.datetime.now():
```
wchamber, line 99: Change condition for ServiceAccountNotUpdatedException to raise that function if there is no expired_at message, i.e. it's still locked, instead of only if the updated account does not have a valid expires_at value.
Before, if the update lock expired during the _wait_for_update loop, we could return the FenceServiceAccount without a received key_json, as the update locked FenceServiceAccount does not have any value for expires_at. Now, the expired update locked value raises the exception.

fence_token_storage.py
dvoet: did you verify this is correct? I think part of the problem was that is was unclear which are the exceptions that should be handled

dvoet (approved):

fence_token_storage.py
wchamber: That was my reading of the docs here[0] but I did not try to provoke a transaction error. Looking some more, it sounds pretty ambiguous in the note[1], so I'll go back to `except:`

[0] https://cloud.google.com/appengine/docs/standard/python/datastore/transactions#what_can_be_done_in_a_transaction
[1] https://cloud.google.com/appengine/docs/standard/python/datastore/transactions#using_transactions

gpolumbo-broad:
Since you're doing some rewriting here anyway, I would suggest a rename of `FenceTokenStorage` to something like, `ProviderTokenStorage`.  The reason we have anything named Fence in the code today is because when we were first writing this stuff, there was one and only one known provider, Fence.  Then we got a second one, which was just another instance of Fence, but technically, these Providers don't _need_ to be Fences.  In general in the code, any classes or variables we have named `Fence` we can probably rename to just `Provider` and it will hopefully be a little clearer and less confusing.

fence_token_storage.py
```diff
@@ -0,0 +1,159 @@
...
+        Delete the stored fence service account info for the fsa_key.
+        ":return returns the stored value for the key or None if it did not exist.
+        """
+        fence_service_account = fsa_key.get()
+        if fence_service_account:
+            fsa_key.delete()
+        return fence_service_account.key_json if fence_service_account else None
+
+    def get_or_create(self, fsa_key, prep_key_fn, fence_fetch_fn):
+        """Retrieve the stored for key, waiting as needed, or create and store the value for the key.
```
gpolumbo-broad: `Retrieve the stored for key`

fence_token_storage.py
```diff
@@ -0,0 +1,159 @@
...
+
+        :param prep_key_fn The function to create the input to fence_fetch_fn from 'fsa_key' once we know
+        fence_fetch_fn will be called. This is separated from fence_fetch_fn so that this can be slow but not spend as
+         much time locking.
+        :param fence_fetch_fn: The function to fetch the credentials from the fence. This function will ensure that
+        fence_fetch_fn is not called multiple times concurrently. Arguments should work as
+        fence_fetch_fn(prep_key_fn(key)) -> returns the string fence account credentials.
+        :return returns the fence service account json key and the expiration time for that value.
+        """
```
gpolumbo-broad, line 71: Missing description for `:param fsa_key`

wchamber:
> Since you're doing some rewriting here anyway, I would suggest a rename of `FenceTokenStorage` to something like, `ProviderTokenStorage`...

Greg and I talked offline. s/fence/provider seems good, but we'll do it in a follow up pull so that we can do all of the moving and renaming at once. I've filed CA-657 to track this work.

fence_token_storage.py
```diff
@@ -0,0 +1,159 @@
...
+        Delete the stored fence service account info for the fsa_key.
+        ":return returns the stored value for the key or None if it did not exist.
+        """
+        fence_service_account = fsa_key.get()
+        if fence_service_account:
+            fsa_key.delete()
+        return fence_service_account.key_json if fence_service_account else None
+
+    def get_or_create(self, fsa_key, prep_key_fn, fence_fetch_fn):
```
gpolumbo-broad: I'm not thrilled with this name.  I am not sure I have a better suggestion though either, maybe `get_from_cache_or_provider` or something like that?   Or maybe we could just call it `get` or `retrieve` or something like that.

fence_token_storage.py
```diff
@@ -0,0 +1,159 @@
...
+        fence_service_account = fsa_key.get()
+        now = datetime.datetime.now()
+        if fence_service_account is None or \
+                fence_service_account.expires_at is None or \
+                fence_service_account.expires_at < now:
+            fence_service_account = self._fetch_service_account(fsa_key, prep_key_fn, fence_fetch_fn)
+
+        return (fence_service_account.key_json, fence_service_account.expires_at)
+
+    def _fetch_service_account(self, fsa_key, prep_key_fn, fence_fetch_fn):
```
gpolumbo-broad: I know this is the same name as the previous version, but in the spirit of my other naming change, maybe call this one `_fetch_and_cache_service_account`

fence_token_storage.py
```diff
@@ -0,0 +1,159 @@
+import datetime
+import time
+from google.appengine.ext import ndb
+from google.appengine.api.datastore_errors import TransactionFailedError
+
+# How long to keep a fence service account key before expiring it.
+_FSA_KEY_LIFETIME = datetime.timedelta(days=5)
+
+
+def create_fence_service_account_key(provider_name, user_id):
```
gpolumbo-broad: Perhaps a nit, but I don't believe this actually persists the Key, and when I read the word `create`, I keep expecting this method to make a new object and persist it.  (Old Rails naming conventions).   I might recommend renaming this to `make_fence_service_account_key` or maybe `build_fence_service_account_key`.

gpolumbo-broad (changes requested):

fence_token_storage.py
wchamber: tried to fix up the comment, PTAL

fence_token_storage.py
wchamber: done

fence_token_storage.py
wchamber: Yeah, I'm not wild about the name. Went with `retrieve`, which I like more than `get_or_create`, thanks! (Get sounds like a simple look up to me, and it's more than that. We're at least always retrieving it from somewhere.)

fence_token_storage.py
wchamber: done

fence_token_storage.py
wchamber: Done. I'm not familiar with the Rails convention. In my java land experience, `create` is typical for a factory function, but `build` is pretty reasonable too.

gpolumbo-broad (approved):

PR #92, opened 2020-01-21
[CA-590] json response verification for integration tests

andy7i:

This PR:
addresses: https://broadworkbench.atlassian.net/browse/CA-590 to create robust tests to validate API json responses return the same data between releases and especially for the python2->3 migration/upgrade.

and adds:
- json verification using Json Schema ([reasoning for this is here](https://github.com/DataBiosphere/bond/pull/92/commits/268ebbc5d9ce7a5f5346d1244a92974dd743ceb5)). Any feedback on organizing the json schema would be especially appreciated.
- mock config for local envs

and removes:
- now-duplicated test code

In all cases:

### DIFF
--- a/automation/README.md
+++ b/automation/README.md
@@ -10,3 +10,7 @@ To run the tests from the Bond root directory:
 
 You can optionally specify the environment variable, `BOND_BASE_URL` to specify the host you want to test against.  This
 variable will default to `https://bond-fiab.dsde-dev.broadinstitute.org:31443` unless otherwise specified.
+
+// TODO:  This seems out of date ^ -- 
+`    bond_base_url = "localhost:8080"`
+ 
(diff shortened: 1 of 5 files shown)


andy7i:
jenkins retest

dvoet (approved):
I like the approach of using json schemas to validate responses in tests. I don't like that the success conditions for a test are removed from the test itself. I am also concerned that people think these schemas can be reused from test to test (such a practice will lead to artificially coupled tests).

I am approving the PR because I don't have a solution. The schemas are verbose so I understand keeping them out of the main test code.

gpolumbo-broad (approved):
There are a lot of schemas and I think it's fine the way you have them.  I didn't go through schema by schema to verify them though as there are so many, so I'm pretty much counting on the tests to verify for us.  You might want to try running the integration tests pointing at stage Bond or something to have them run with a real Fence

PR #94, opened 2020-01-24
CA-648 Add a Datastore backed implementation of CacheApi.

wchamber:
To be used to replace memcache implementation. Also needs work to expose an endpoint to hit with a cron to periodically delete expired entries.

In all cases:

### DIFF
--- a/datastore_cache_api.py
+++ b/datastore_cache_api.py
@@ -1,0 +1,62 @@
+import datetime
+from google.appengine.ext import ndb
+from cache_api import CacheApi
+import logging
+
+_NO_EXPIRATION_DATETIME = datetime.datetime(year=3000, month=1, day=1)
+
+
+class CacheEntry(ndb.Model):
+    """Datastore model for cache values and expirations."""
+    # The value stored in the cache.
+    value = ndb.PickleProperty()
+    # The datetime when to expire the cache entry, or _NO_EXPIRATION_DATETIME for no expiration.
+    # Datastore docs warn against doing this at high write rates.
+    # https://cloud.google.com/datastore/docs/best-practices#deletions
+    expires_at = ndb.DateTimeProperty()
+
+
+class DatastoreCacheApi(CacheApi):
+    """
+    A CacheApi backed by Datastore.
+
+    N.B. expiration does not happen automatically. In appengine we use a cron.yml to ensure that we periodically
+    delete expired entries.
+    """
+
+    def add(self, key, value, expires_in=0, namespace=None):
+        CacheEntry(key=DatastoreCacheApi._build_cache_key(key, namespace), value=value,
+                   expires_at=DatastoreCacheApi._calculate_expiration(expires_in)).put()
+        return True
+
+    def get(self, key, namespace=None):
+        entry = DatastoreCacheApi._build_cache_key(key, namespace).get()
+        if not entry or entry.expires_at < datetime.datetime.now():
+            return None
+        return entry.value
+
+    @staticmethod
+    def delete_expired_entries():
+        """Deletes the entries that have expired. This must be done periodically. """
+        expired_entries = CacheEntry.query(CacheEntry.expires_at < datetime.datetime.now())
+        deletions = ndb.delete_multi([key for key in expired_entries.iter(keys_only=True)])
+        logging.info("Deleted %d cache entries.", len(deletions))
+
+    @staticmethod
+    def _build_cache_key(key, namespace):
+        """Create an ndb Key for the key and namespace."""
+        if namespace is not None:
+            return ndb.Key("cache namespace", namespace, CacheEntry, key)
+        else:
+            return ndb.Key(CacheEntry, key)
+
+    @staticmethod
+    def _calculate_expiration(expires_in):
+        """
+        Calculates the datetime that an entry should expire.
+        :param expires_in in how many seconds to expire the key, or 0 if it should not expire.
+        :return: The expiration datetime or else None if it should not expire.
+        """
+        if expires_in == 0:
+            return _NO_EXPIRATION_DATETIME
+        return datetime.datetime.now() + datetime.timedelta(seconds=expires_in)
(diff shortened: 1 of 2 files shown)


wchamber:
jenkins retest plz

datastore_cache_api.py
```diff
@@ -0,0 +1,62 @@
+import datetime
+from google.appengine.ext import ndb
+from cache_api import CacheApi
+import logging
+
+_NO_EXPIRATION_DATETIME = datetime.datetime(year=3000, month=1, day=1)
```
andy7i, line 6: Y3K might hit Bond hard /s

datastore_cache_api.py
```diff
@@ -0,0 +1,62 @@
+import datetime
+from google.appengine.ext import ndb
```
andy7i, line 2: I think appengine's ndb doesn't support python 3, whereas python-ndb supports both python 2 and 3. If both of those statements are true, then maybe it would be worth switching to python-ndb? Not sure how much work / how feasible that is though.

datastore_cache_api.py
```diff
@@ -0,0 +1,62 @@
...
+class DatastoreCacheApi(CacheApi):
+    """
+    A CacheApi backed by Datastore.
+
+    N.B. expiration does not happen automatically. In appengine we use a cron.yml to ensure that we periodically
+    delete expired entries.
+    """
+
+    def add(self, key, value, expires_in=0, namespace=None):
+        CacheEntry(key=DatastoreCacheApi._build_cache_key(key, namespace), value=value,
```
andy7i, line 28: `ndb.Model` in `CacheEntry(ndb.Model)` was a bit confusing for me but I think I sort of get it. Seems like CacheEntry automatically sets the key of the entity. 

>You cannot easily define a property named "key", "id", "parent", or "namespace". If you pass, for example, key="foo" in a constructor or populate() call, it sets the entity's key, not a property attribute named "key".

https://cloud.google.com/appengine/docs/standard/python/ndb/modelclass

tests/unit/datastore_cache_api_test.py
```diff
@@ -0,0 +1,44 @@
...
+from google.appengine.ext import ndb
+from google.appengine.ext import testbed
+from tests.unit import cache_api_test
+
+
+class DatstoreCacheApiTestCase(unittest.TestCase, cache_api_test.CacheApiTest):
+    def setUp(self):
+        self.testbed = testbed.Testbed()
+        self.testbed.activate()
+        self.testbed.init_memcache_stub()
```
andy7i, line 15: Does this test still need memcache? (same thing on L19)

datastore_cache_api.py
```diff
@@ -0,0 +1,62 @@
...
+import logging
+
+_NO_EXPIRATION_DATETIME = datetime.datetime(year=3000, month=1, day=1)
+
+
+class CacheEntry(ndb.Model):
+    """Datastore model for cache values and expirations."""
+    # The value stored in the cache.
+    value = ndb.PickleProperty()
+    # The datetime when to expire the cache entry, or _NO_EXPIRATION_DATETIME for no expiration.
```
andy7i, line 13: Maybe I don't understand caches well but it seems safer to me to set a short default expiration value instead of the default being an unlimited expiration date.

datastore_cache_api.py
```diff
@@ -0,0 +1,62 @@
+import datetime
```
andy7i, line 1: Is this file replacing one of the existing *cache_api.py files? It would be easier for me to review these files while comparing it to the existing files, but I also understand if it wasn't included in order to keep the PR size small. Same goes for the test files.

andy7i (approved):
Mostly just questions -- nothing blocking a merge.

Test seems good -- did not run it.

datastore_cache_api.py
wchamber: Yes. I'm switching to Cloud NDB (is that the same as python-ndb) as a part of this ticket. While not the concern of this PR, I need a whole bunch of things to be in place to switch NDB libraries. Yaks all the way down. See CA-648 for some more details.

datastore_cache_api.py
wchamber: I agree that that makes more sense, but this matches the default behavior of the memcache api. Wasn't trying to change behavior in this PR, but can if wanted.

datastore_cache_api.py
wchamber: Yes, this will replace the memcache_api.py file (and the corresponding memcache_api_test.py). Unfortunately, I need to slice it between PRs some manner and I had Greg review the other PR.

datastore_cache_api.py
wchamber: Going to give my best explanation, sorry if I'm repeating things you understood.
CacheEntry subclassing ndb.Model gives it the hooks to act as an ORM-like thingy. ndb.Model subclasses look for attributes that are special ndb.*Properties. These get persisted and retrieved on calls to `.put()` and `.get()`.
Datastore/ndb is a key-value store. By default, ndb gives things automatically generated unique keys. But we want to be able to find logically equivalent items, so we create meaningful keys, using `_build_cache_key`. Then, `CacheEntry(key=_build_cache_key...`  tells ndb to use the special "key" parameter as the key for datastore key-value lookup (with CacheEntry being the value).
Happy to talk more, or others who have been touching datastore for longer can chime in.

tests/unit/datastore_cache_api_test.py
wchamber: Yeah, datastore uses it by default. Added a comment.

wchamber:
Jenkins is failing because of an issue at head in `develop`, see PR #99. (Well, I assume that because I can't actually see Jenkins results.)

wchamber:
I'm changing this to merge into the `python3-feature` branch instead of `develop`. We want a URL route wired up to clear expired datastore values, and it seems simpler to do that with the new flask stuff instead of setting up it up in the old endpoints and then moving that to the new flask stuff. OTOH, this means it's a nother change when we do the release to python3, making things less incremental. I could be persuaded.

datastore_cache_api.py
dvoet: don't change behavior

dvoet (approved):

PR #97, opened 2020-01-27
CA-648 Use Datastore emulator for testing

wchamber:
Set up some integration tests using the Datastore emulator for testing interactions with Datastore. Hook these up to be run by circleci. Add instructions for running datastore emulator tests locally.

With the move to python3, the testbed datastore stub is being removed so we need to use the emulator for testing.

In all cases:


tests/datastore_emulator/README.md
```diff
@@ -0,0 +1,24 @@
...
+These are tests that require Datastore to run. The Datastore Emulator allows us to run these tests with some environment
+modifications.
+
+See https://cloud.google.com/datastore/docs/tools/datastore-emulator
+
+## Running Locally
+### One time setup
+Do the one time setup of installing the Datastore Emulator.
+
+`gcloud components install cloud-datastore-emulator`
```
gpolumbo-broad, line 11: Is there are reason why you recommend using `gcloud components install` here, but in the circle config you use `apt-get` to install the emulator?

gpolumbo-broad (approved):

tests/datastore_emulator/README.md
wchamber: Good question. The website linked above uses the `gcloud` command. My Google laptop barfed and insisted I used apt-get. circleci does the same thing. Since the `gcloud components install` points at the apt-get, it seemed reasonable since Google was recommending it.

WDYT?

sample error:
```
ERROR: (gcloud.components.install) 
You cannot perform this action because the Cloud SDK component manager 
is disabled for this installation. You can run the following command 
to achieve the same result for this installation: 

sudo apt-get install google-cloud-sdk-datastore-emulator
```

dvoet (approved):

tests/datastore_emulator/README.md
gpolumbo-broad: Ah, ok, that's fine.  It is probably a function of the base image we're using.  There's a number of different base images we could use and there's probably one that already has the emulator installed on it.  Image might be larger though which would probably just mean our build runs slower.  Either way is fine.  

In any case, thanks for the explanation, I'm not worried about it since there's a reasonable explanation.

PR #102, opened 2020-01-28
CA-648 Move datastore_cache_api_test to datastore_emulators directory.

wchamber:
Move datastore_cache_api_test to datastore_emulators directory.
Add assertions to CacheApiTest to verify that there are no previous entries at the start of a test.
Add utilities for setting up and clearing out datastore emulator properly for tests.
Add shell script for starting the Datastore Emulator for testing and use it in circleci.

In all cases:


tests/datastore_emulator/datastore_cache_api_test.py
```diff
@@ -2,23 +2,16 @@
 import unittest
```
andy7i, line 2: `tests/unit/datastore_cache_api_test.py → tests/datastore_emulator/datastore_cache_api_test.py`

Are there instructions anywhere for how to run these tests?
Also for consistency, the readme says we should put tests in `tests/unit` directory and `tests/integration` directory.

andy7i (approved):
I like the abstraction of the setup / teardown. One comment but otherwise LGTM 👍

tests/datastore_emulator/datastore_cache_api_test.py
wchamber: Thanks, good point. There are instructions in this directory, but I added a note to the top level README to point to the `tests/datastore_emulator/README` following your style in commit 9188b7716.

I think it's worth keeping them in a separate directory because they require different commands to set up than the rest of the integration tests.
(Thinking about this a little more, I made a simple shell script to start the emulator and referred to it in the README and the circleci config, thanks!)

dvoet (approved):

PR #103, opened 2020-01-28
CA-648 Change Datastore client library to Cloud NDB

wchamber:
Change which Datastore client library is used in bond to Cloud NDB. Cloud NDB is supported in both python 2 & 3, which will help us migrate.

Modify requirements and flask app set up.
Change imports of tests.unit module as there seems to be some module naming conflict now.
Modify datastore emulator tests to work with new client library.

In all cases:


dvoet (approved):

tests/unit/memcache_api_test.py
```diff
@@ -14,5 +13,4 @@ def setUp(self):
         self.setUpCache(memcache_api.MemcacheApi())
 
     def tearDown(self):
```
gpolumbo-broad, line 15: Can we just delete this file?

gpolumbo-broad (approved):

tests/unit/memcache_api_test.py
wchamber: Yes. I started this PR before we talked about switching over to datastore cache usage at the same time as the rest of python3 feature branch. Now there's no need to keep both. Will follow up in another PR.

wchamber:
Per offline talk, I expect this to break the jenkins test in the feature branch due to problems starting up dev_appserver with dependencies vs 3rd party libraries. We're going to go ahead and try to fix this forward in python3 where dependencies work differently again.

PR #104, opened 2020-01-29
[CA-592] Throw JSON exceptions instead of HTML ones

marctalbott:
This PR is mainly aimed at re-implementing the default HTML exceptions that Flask was throwing as JSON exceptions so that they would be (nearly) identical to the errors that we threw when we used endpoints. This is part of CA-592 since it was initially left off of the work done for that ticket, but was necessary for tests to pass.

In all cases:

### DIFF
--- a/automation/helpers/json_responses.py
+++ b/automation/helpers/json_responses.py
@@ -44,7 +44,7 @@ json_schema_test_status = {
             "enum": [
               "fence",
               "dcf-fence",
-              "memcache",
+              "cache",
               "datastore",
               "sam"
             ],
@@ -127,7 +127,7 @@ json_schema_test_get_auth_url_without_params = {
               "message": {
                 "type": "string",
                 "enum": [
-                  "Error parsing ProtoRPC request (Unable to parse request content: Message CombinedContainer is missing required field redirect_uri)"
+                  "The browser (or proxy) sent a request that this server could not understand."
                 ],
               },
               "reason": {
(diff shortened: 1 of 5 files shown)


dvoet (approved):

automation/helpers/json_responses.py
```diff
@@ -548,7 +548,7 @@
               "message": {
                 "type": "string",
                 "enum": [
-                  "Error parsing ProtoRPC request (Unable to parse request content: Message CombinedContainer is missing required field oauthcode)"
+                  "The browser (or proxy) sent a request that this server could not understand."
```
andy7i, line 551: With the new error messages, it seems like we're losing informative error messaging which is too bad. I understand if that's outside the scope of this ticket though

json_exception_handler.py
```diff
@@ -0,0 +1,36 @@
+from flask import jsonify
+from werkzeug import exceptions
+
+
+class JsonExceptionHandler(object):
+    def __init__(self, app=None):
+        if app:
+            self.init_app(app)
+
+    def json_dict(self, status_code, message):
```
andy7i, line 10: What does `json_dict` do if the status code is not one of the codes listed? Should we allow that to pass through?

andy7i (commented):
drive by comments

gpolumbo-broad (approved):

automation/helpers/json_responses.py
marctalbott: Yeah, unfortunately trying to get Flask and webargs to re-generate the same error message that endpoints and protorpc were generating was too much of a headache so I left it like this. [CA-500](https://broadworkbench.atlassian.net/browse/CA-500) should address this.

PR #111, opened 2020-02-03
Checkpoint getting running flask app locally with ./run_local.sh 

wchamber:
(with datastore emulator already running).
Moved some BOND_ACCEPTED environment variables in app.yaml.ctmpl to be provided by config.ini.ctmpl.
Use local credentials when datastore emulator host is set.

In all cases:


bond_app/authentication.py
```diff
@@ -48,12 +48,6 @@ def __init__(self, accepted_audience_prefixes, accepted_email_suffixes, max_toke
         self.max_token_life = max_token_life
 
 
-def default_config():
-    return AuthenticationConfig(os.environ['BOND_ACCEPTED_AUDIENCE_PREFIXES'].split(),
```
marctalbott, line 52: let's remove the os import in this file since I don't think it was being used other than here

marctalbott (approved):
Just the one nit. I'm also going to kick off an automation test run against this branch

bond_app/authentication.py
wchamber: done

PR #114, opened 2020-02-04
Update README for running bond in python3

wchamber:
Remove instructions on setting up paths for dev_appserver, as its no longer needed.
Remove notes on nose as running tests with unittest is straightforward now.
Add some more notes on setting up a python2 env to run the datastore emulator.
Remove scratch notes.

In all cases:


.python-version
```diff
@@ -1 +0,0 @@
-2.7.15
```
marctalbott: what was this used by? do we need to update it?

marctalbott (approved):

.python-version
wchamber: Oops, meant to not touch this. I'll revert.
Greg added it recently in fad0587c9b. It's not obvious to me that we need it.

PR #120, opened 2020-02-05
[No Ticket] add config option to ignore certs for a fence provider

marctalbott:
We realized when we deployed our python 3 flask app to dev that staging.datastage.io's certificate is not valid. The old library that we used to use to talk to the provider (`urlfetch`) did not do any cert validation. However, the new library that we use (`requests`) does, and it was failing since the certificate was invalid. This PR adds a config option to not do any cert validation which serves as a workaround while we wait for datastage to fix their staging site.

In all cases:

### DIFF
--- a/bond_app/routes.py
+++ b/bond_app/routes.py
@@ -228,7 +228,9 @@ def clear_expired_datastore_entries():
 def get_status():
     sam_base_url = config.get('sam', 'BASE_URL')
 
-    providers = {section_name: FenceApi(config.get(section_name, 'FENCE_BASE_URL'))
+    providers = {section_name: FenceApi(config.get(section_name, 'FENCE_BASE_URL'),
+                                        config.get(section_name, 'IGNORE_CERTIFICATES')
+                                        if config.has_option(section_name, 'IGNORE_CERTIFICATES') else False)
                  for section_name in config.sections() if is_provider(section_name)}
 
     sam_api = SamApi(sam_base_url)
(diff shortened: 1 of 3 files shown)


bond_app/routes.py
```diff
@@ -228,7 +228,9 @@ def clear_expired_datastore_entries():
 def get_status():
     sam_base_url = config.get('sam', 'BASE_URL')
 
-    providers = {section_name: FenceApi(config.get(section_name, 'FENCE_BASE_URL'))
+    providers = {section_name: FenceApi(config.get(section_name, 'FENCE_BASE_URL'),
+                                        config.get(section_name, 'IGNORE_CERTIFICATES')
+                                        if config.has_option(section_name, 'IGNORE_CERTIFICATES') else False)
```
wchamber, line 233: qq, does config.has_option retrieve the value if its present, or only check for the existence of the option?
e.g. if we have IGNORE_CERTIFICATES specified to False, will has_option still return true?

wchamber (approved):
lgtm, just checking on config.has_option

dvoet (approved):

bond_app/routes.py
marctalbott: It only checks for the existence of the option. So if IGNORE_CERTIFICATES is False, `config.has_option` will return True, and we'll pass along the specified config value

wchamber: Maybe not a big deal, but I think that's pretty confusing. I'd expect setting IGNORE_CERTIFICATES to False to do that. I have a mild preference for fixing this up if we're going to keep this code after the fence fixes their certs.

PR #122, opened 2020-02-07
CA-500 Improve Bond error codes and logging

wchamber:
Change Bond's error for unable to find refresh tokens to 404 instead of 400. Suggest to users that they relink their accounts to Bond.
Add some logging statements in odder corner cases of the code.

In all cases:


wchamber:
I'm opening this for review now as merging into `python3-feature`, but I think we should wait to merge this into develop until we've already successfully deployed py3 Bond.
I looked at all of the places that we're raising exceptions in Bond.

bond_app/fence_token_vending.py
```diff
@@ -62,15 +62,9 @@ def get_service_account_key_json(self, user_info):
...
-        real_user_info = self.sam_api.user_info(user_info.token)
-        if real_user_info is None:
-            raise exceptions.Unauthorized("user not found in sam")
-        return real_user_info
-
     def _get_oauth_access_token(self, provider_user):
         refresh_token = self.refresh_token_store.lookup(provider_user.user_id, self.provider_name)
         if refresh_token is None:
-            raise exceptions.BadRequest("Fence account not linked")
+            raise exceptions.NotFound("Fence account not linked. {}".format(str(provider_user)))
```
dvoet, line 73: What is the rationale for this change? I don't recall what we were thinking at the time but BadRequest must have made sense.

bond_app/sam_api.py
```diff
@@ -12,14 +12,14 @@ def user_info(self, access_token):
...
-        :return: dict with userSubjectId and userEmail keys or None if user does not exist in sam
+        :return: dict with userSubjectId and userEmail keys or else raises an exception if user does not exist in sam
         """
         headers = {'Authorization': 'Bearer ' + access_token}
         result = requests.get(url=self.base_url + '/register/user/v1?userDetailsOnly=true', headers=headers)
         if result.status_code == 200:
             return json.loads(result.content)["userInfo"]
         elif result.status_code == 404:
-            return None
+            raise exceptions.Unauthorized("user not found in sam")
```
dvoet: I disagree with this change. The sam_api module should reflect the sam api and let the caller determine what to do with the response. The appropriate return value for a 404 is None and the caller should raise the appropriate exception or handle it

dvoet (changes requested):

bond_app/sam_api.py
gpolumbo-broad: That being said, if the status is anything other than `2xx` here, we should add a logging statement to log the response from Sam.

gpolumbo-broad (approved):

bond_app/sam_api.py
wchamber: sg, reverted fence_token_vending and added checks to bond.py to handle None return case.
Also added logging here.

bond_app/fence_token_vending.py
wchamber: I don't think BadRequest is a good match for what's going on. Bond wasn't apple to find a valid refresh token for the user, either because it expired or the user has never linked their account. There's nothing bad about the user's request. They wouldn't have to change the request if their token existed, so I don't think BadRequest is a good fit.

I'm not sure if Unauthorized, Forbidden, or NotFound is best. We didn't find the refresh token, we're not sure why. The user might just need to authorize their fence account through Bond, which sounds Unauthorized-ish to me. Or maybe the current action is Forbidden since we know who the user is in Sam, but they don't have a refresh token that matches. WDYT?

gpolumbo-broad (approved):

dvoet (approved):

PR #124, opened 2020-02-11
CA-533 Remove pycrypto dependency

wchamber:
With the move to python3 GAE, we no longer seem to need our pycrypto dependency.

In all cases:


wchamber:
I'm going to let the python3-feature to develop merge and deployment go through before merging this in. Could go ahead and do it now if people like instead. My only concern is that there may be problems that only show up in a GAE deployment, as opposed to all of our local testing, though I don't think that's the case anymore with GAE python3.

gpolumbo-broad (approved):

dvoet (approved):

PR #138, opened 2020-02-26
[No ticket] Fix broken status logs

marctalbott:
In order to fix the content type response stuff, we had to start returning a tuple in our responses, including the status response. However, when the status was not ok, we would log that status and expect the status to be json. The json that we were logging was changed to a tuple for content type, but that broke our status logging. This PR now grabs the first element out of the tuple, which is the json we want, and logs that when Bond's status is not ok.

In all cases:

### DIFF
--- a/bond_app/routes.py
+++ b/bond_app/routes.py
@@ -251,5 +251,5 @@ def get_status():
     if ok:
         return response
     else:
-        logging.warning("Bond status NOT OK:\n%s" % response)
-        raise exceptions.InternalServerError(response)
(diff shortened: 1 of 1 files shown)


yonghaoy (approved):

dvoet (approved):
is this the only place?

marctalbott:
Yes, everywhere else `json_response` is used it is returned directly and Flask handles it. This was the only place we were storing it in a variable and doing anything with the value.

PR #142, opened 2020-03-23
Add a README note about rendering configs before running docker compose.

wchamber:

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,9 @@ You can also check [http://0.0.0.0:8432](http://0.0.0.0:8432) which should show 
 
 ## Run in a Docker container
 
+First, render configs (see above). If you try to run the docker compose before rendering, subsequent rendering may not
+work. Try removing created directories and re-rendering.
+
 Choose one of the options below:
 
 a) To run an existing image:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,6 +98,9 @@ You can also check [http://0.0.0.0:8432](http://0.0.0.0:8432) which should show
 
 ## Run in a Docker container
 
+First, render configs (see above). If you try to run the docker compose before rendering, subsequent rendering may not
+work. Try removing created directories and re-rendering.
```
gpolumbo-broad: Do you have an example of what the `created directories` might be and how did they get created and why can/should they be deleted?

rsasch (changes requested):
I think it might be clearest if you number the steps under "Running locally" so that "Render configs" is 1, "Run on your local environment" is 2a, and "Run in a Docker container" is 2b (or something similar.   -- the two docker options will need to change to I and II or something, but you get the idea....)

README.md
wchamber: added some more info, ptal

wchamber:
> I think it might be clearest if you number the steps under "Running locally" so that "Render configs" is 1, "Run on your local environment" is 2a, and "Run in a Docker container" is 2b (or something similar. -- the two docker options will need to change to I and II or something, but you get the idea....)

Thanks! I added it as step 1 for both options and left the explanation above. That seemed the simplest with markdown auto rendering numbers.

rsasch (approved):
LGTM 👍

gpolumbo-broad (approved):

PR #152, opened 2020-07-23
[CA-943] Tolerate unlinking expired refresh tokens

marctalbott:
Ticket: [CA-943](https://broadworkbench.atlassian.net/browse/CA-943)
When we try to revoke a refresh token that has already expired, `fence` responds with a 400. This PR allows Bond to tolerate that 400 and continue and delete the refresh token from the token store.

Still need to figure out how best to test both this and unlinking more generally.

In all cases:

### DIFF
--- a/bond_app/fence_api.py
+++ b/bond_app/fence_api.py
@@ -37,7 +37,9 @@ class FenceApi:
     def revoke_refresh_token(self, refresh_token):
         result = requests.post(url=self.revoke_url, data=refresh_token)
         if result.status_code // 100 != 2:
-            raise exceptions.InternalServerError("fence status code {}, error body {}".format(result.status_code, result.content))
+            if result.status_code != 400:
+                raise exceptions.InternalServerError("fence status code {}, error body {}"
+                                                     .format(result.status_code, result.content))
 
     def status(self):
         """
--- a/bond_app/oauth_adapter.py
+++ b/bond_app/oauth_adapter.py
@@ -1,5 +1,6 @@
 import base64
 from werkzeug import exceptions
+import logging
 
 import requests
 from requests.auth import HTTPBasicAuth
@@ -70,6 +71,7 @@ class OauthAdapter:
                                data={"token": refresh_token},
                                headers={"Authorization": "Basic %s" % base64.b64encode(
                                    "{}:{}".format(self.client_id, self.client_secret).encode()).decode()})
+        logging.info("POST {} - status code: {}".format(revoke_url, result.status_code))
         if result.status_code // 100 != 2:
             # If the refresh token has already expired, the auth provider will return a 400 when we try to revoke it.
             # We don't want to fail if that happens, we just want to keep going and delete the expired token on our end
--- a/bond_app/oauth_adapter.py
+++ b/bond_app/oauth_adapter.py
@@ -71,7 +71,7 @@ class OauthAdapter:
                                data={"token": refresh_token},
                                headers={"Authorization": "Basic %s" % base64.b64encode(
                                    "{}:{}".format(self.client_id, self.client_secret).encode()).decode()})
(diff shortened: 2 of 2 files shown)


bond_app/oauth_adapter.py
```diff
@@ -71,5 +71,8 @@ def revoke_refresh_token(self, refresh_token):
                                headers={"Authorization": "Basic %s" % base64.b64encode(
                                    "{}:{}".format(self.client_id, self.client_secret).encode()).decode()})
         if result.status_code // 100 != 2:
-            raise exceptions.InternalServerError("revoke url {}, status code {}, error body {}".
-                                                 format(revoke_url, result.status_code, result.content))
+            # If the refresh token has already expired, the auth provider will return a 400 when we try to revoke it.
+            # We don't want to fail if that happens, we just want to keep going and delete the expired token on our end
+            if result.status_code != 400:
```
gpolumbo-broad, line 78: Couple of suggestions here.

1. I feel like we should log more in Bond in general.  I'm inclined to log all responses from Fence, even 2xx's.  At the very least, I suggest we log the 4xx response here without raising an exception because it's sort of "not what we expected, but we're going to pretend that everything is fine"
2.  Should we also "succeed" if we get back a 5xx from Fence?  We should be the only ones with the Refresh Token and if we destroy our record of it, we're effectively doing what the user wants from our side.  That said, a 4xx or 5xx indicates that the token was not successfully revoked/destroyed on the Fence side of things, but I don't know what that means?  What does an OAuth provider do when you revoke a Refresh Token?  I think all that this would mean is that Fence may still indicate that "Terra has token with these scopes" if we fail to properly revoke the token.  That's not ideal, but also maybe not a big problem?

bond_app/fence_api.py
```diff
@@ -37,7 +37,9 @@ def delete_credentials_google(self, access_token, key_id):
     def revoke_refresh_token(self, refresh_token):
```
gpolumbo-broad, line 37: Looking at the code, I'm not sure we ever call this method.  I think we only ever use the implementation in `OauthAdapter`

marctalbott: Yeah, I changed it here for consistency but couldn't actually figure out what `fence_api.py` was for or whether I needed to change this function at all

bond_app/oauth_adapter.py
marctalbott: with regards to 1. I added a log message for the response we get back from fence when we revoke a refresh token. I'm not sure how much more logging to try to add as overhauling logging in all of Bond feels out of scope for this ticket, but I do agree that the logging in Bond is still lacking.

marctalbott: As for 2. I was reading the [specification for the revoke url](https://tools.ietf.org/html/rfc7009) to try and answer some of your questions here. I think what I found actually means we shouldn't be tolerant of 400 responses here after all. 

> The authorization server responds with HTTP status code 200 if the
   token has been revoked successfully or if the client submitted an
   invalid token.
Note: invalid tokens do not cause an error response since the client
   cannot handle such an error in a reasonable way.  Moreover, the
   purpose of the revocation request, invalidating the particular token,
   is already achieved.

Based on this, fence shouldn't actually be responding with a 400 when we try to revoke an invalid token.

marctalbott:
Technically, UChicago should return a 200 when we try to unlink an expired refresh token, so they are going to make that change. In the meantime, we will tolerate the 400 and also log all responses from providers when revoking a refresh token.

gpolumbo-broad (approved):

marctalbott:
Tested this manually by deploying it to dev. Verified that I could delete an expired token without Bond throwing errors. Created a ticket to track the work to remove this change once Fence fixes their revoke URL https://broadworkbench.atlassian.net/browse/CA-1021

PR #153, opened 2020-08-24
[CA-985] delete SA key from cache when removing service account while unlinking

marctalbott:
Ticket: [CA-985](https://broadworkbench.atlassian.net/browse/CA-985)
We used to just let deleted keys for unlinked accounts hang around in the cache until they expired and were cleaned up by the cron job, but this PR changes that so that we will clear any keys out when an account is unlinked. This prevents the behavior described in the ticket where a user can unlink using Bond and continue to get the cached SA key. Although that SA key should be invalid after we delete it with Fence while unlinking, it is still good practice to clear it out of the cache when we are deleting all other credentials for the user.

In all cases:

### DIFF
--- a/bond_app/datastore_cache_api.py
+++ b/bond_app/datastore_cache_api.py
@@ -35,6 +35,9 @@ class DatastoreCacheApi(CacheApi):
             return None
         return entry.value
 
+    def delete(self, key, namespace=None):
+        DatastoreCacheApi._build_cache_key(key, namespace).delete()
+
     @staticmethod
     def delete_expired_entries():
         """Deletes the entries that have expired. This must be done periodically. """
--- a/tests/unit/cache_api_test.py
+++ b/tests/unit/cache_api_test.py
@@ -41,6 +41,23 @@ class CacheApiTest(object):
         self.assertIsNone(self.cache.get('foo'))
         self.assertIsNone(self.cache.get('foo', namespace='bar'))
 
+    def test_simple_delete(self):
+        self.assertTrue(self.cache.add('foo', 42))
+        self.assertEqual(self.cache.get('foo'), 42)
+        self.assertIsNone(self.cache.delete('foo'))
+        self.assertIsNone(self.cache.get('foo'))
+
+    def test_delete_with_namespace(self):
+        self.assertTrue(self.cache.add('bar', 11, namespace='baz'))
+        self.assertIsNone(self.cache.get('bar'))
+        self.assertEqual(self.cache.get('bar', namespace='baz'), 11)
+        self.assertIsNone(self.cache.delete('bar', namespace='baz'))
+        self.assertIsNone(self.cache.get('bar', namespace='baz'))
+
+    def test_delete_tolerates_missing_key(self):
+        self.assertIsNone(self.cache.delete('foo'))
+        self.assertIsNone(self.cache.delete('foo', namespace='bar'))
+
 
 class FakeCacheApiTestCase(unittest.TestCase, CacheApiTest):
     def setUp(self):
--- a/tests/unit/fake_cache_api.py
+++ b/tests/unit/fake_cache_api.py
@@ -46,3 +46,12 @@ class FakeCacheApi(CacheApi):
         else:
             timed_value = self.cache.get(key)
         return timed_value.get_valid_value() if timed_value else None
+
+    def delete(self, key, namespace=None):
+        if namespace:
+            namespace_dict = self.namespaces.get(namespace)
+            if not namespace_dict:
+                return None
+            namespace_dict.pop(key, None)
+        else:
+            self.cache.pop(key, None)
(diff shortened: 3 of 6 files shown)


bond_app/datastore_cache_api.py
```diff
@@ -35,6 +35,9 @@ def get(self, key, namespace=None):
             return None
         return entry.value
 
+    def delete(self, key, namespace=None):
+        DatastoreCacheApi._build_cache_key(key, namespace).delete()
```
gpolumbo-broad, line 39: I took a look at https://cloud.google.com/appengine/docs/standard/python/ndb/creating-entities#Deleting_entities which has the following statement:

> Note that this is an operation on the key, not on the entity itself. It always returns None.
 
I am trying to wrap my head around what this means for us.  After we call `delete`, the entity should no longer have a key, and no key means we shouldn't be able to find the entity using the key in the future.  But does it also mean the entities just hang around forever without a key in datastore?  Even our cron job looks like it's retrieving entities using a non-key based query, but then it tries to delete only those with keys:

https://github.com/DataBiosphere/bond/pull/153/files#diff-3c6e0e052797e6eeaa77a5d69d52f4e1R45

So I don't know what this means.  I don't see a way to delete entities whether they have a key or not!  But I also can't really tell if it matters.  Do we need to update the cache entry first and null out all the fields and then call `delete`?  From a practical and functional perspective I think no.  Not sure if there's a compliance catch here where we need to guarantee that it gets wiped when a user unlinks.

I'm also concerned about the fact that the docs say `It always returns None.`  Do they mean always always?  Like even if it errors or times out or whatever?  I may be paranoid here, but I'm wondering if we shouldn't check that the delete worked here by doing a subsequent `get` to confirm that the the thing is actually gone?

tests/unit/cache_api_test.py
```diff
@@ -41,6 +41,23 @@ def test_expiration(self):
         self.assertIsNone(self.cache.get('foo'))
         self.assertIsNone(self.cache.get('foo', namespace='bar'))
```
andy7i, line 43: Does datastore overwrite existing key/value pairs? I would imagine it does and the behavior would look like this:

```suggestion

    def test_multiple_adds_single_delete(self):
        self.assertTrue(self.cache.add('foo', 0))
        self.assertTrue(self.cache.add('foo', 42))
        self.assertEqual(self.cache.get('foo'), 42)
        self.assertIsNone(self.cache.delete('foo'))
        self.assertIsNone(self.cache.get('foo'))

```

(We don't need to test the behavior of datastore right?)

tests/unit/fake_cache_api.py
```diff
@@ -45,4 +45,13 @@ def get(self, key, namespace=None):
...
+        return timed_value.get_valid_value() if timed_value else None
+
+    def delete(self, key, namespace=None):
+        if namespace:
+            namespace_dict = self.namespaces.get(namespace)
+            if not namespace_dict:
+                return None
+            namespace_dict.pop(key, None)
+        else:
+            self.cache.pop(key, None)
```
andy7i, line 57: what is the `None` for?

andy7i (approved):
👍 ; couple Qs

tests/unit/fake_cache_api.py
marctalbott: The `None` is the default value. If there is no value for the key in the cache, it'll return `None` instead of throwing an exception. This imitates the behavior of the real cache which returns `None` regardless of if it deletes anything

tests/unit/cache_api_test.py
marctalbott: I'm not sure what datastore does -- is there something you're concerned about that's impacted by whether or not it overwrites existing values? I don't think we necessarily need this particular test, but I think I'm going to try and add an an unlinking test that's backed by the datastore emulator

andy7i: yeah I was just wondering if there's any reason to believe that the above test would fail (like if datastore stored both of the `add` entries instead of overwriting the existing one... and I just realized that this is probably using fake_cache_api.py. Having a test using the datastore emulator sounds like a good idea!

bond_app/datastore_cache_api.py
gpolumbo-broad: Maybe the answer is that we just leave a comment here with a link to the docs and maybe copy/paste some of the text here, just in case google modifies it or takes it away.  We can just leave a note to our future selves saying that Google's documentation was not clear about actually gets "deleted" when we call delete.  Some sort of breadcrumb to our future selves that if this delete seems to not be fully deleting things.

marctalbott: I think it just means you call `.delete()` on the key object, not the entity object. I don't think we need to worry about entities without keys. According to the Datastore documentation, all entities have a unique key https://cloud.google.com/datastore/docs/concepts/entities. 

As for it always returning `None`, I think it just means whether or not it finds the key you're trying to delete. I'd think that any exceptions would still be raised, but I can look into it.

marctalbott: I've tested it with a datastore emulator and confirmed that the changes in this PR do indeed clear out the cached SA key from datastore

gpolumbo-broad (approved):

PR #155, opened 2020-09-23
[CA-1030] Cache Sam user info

marctalbott:
Ticket: [CA-1030](https://broadworkbench.atlassian.net/browse/CA-1030)

Two main things in this PR:
1. We used to cache a user's SA key for pets. We did this because during the course of a workflow, there can be a lot of requests to Bond as the pet asking for the SA key that belongs to the user. Bond needs to ask Sam who the pet belongs to in order to answer this question, but we didn't want to have to hit Sam on every request, so we cached the SA key and associated it with the pet directly. Unfortunately, this made it difficult to properly clear out a user's information when they unlinked because an invalidated SA key was being cached in reference to a user's pet -- a relationship that we do not know when unlinking a user. This PR changes it so that we do *not* cache the SA key for a user's pet. To avoid overwhelming Sam with requests, this PR adds a cache which stores the response from Sam when Bond asks for the user information on an incoming request. This allows us to determine which user a pet belongs to on every request to Bond without slamming Sam.

2. I reworked how we pass around user info in Bond. Previously, we passed `UserInfo` objects which contained information about the token for a request that we'd obtained from Google. The `UserInfo` object was only ever used to grab user information from Sam, so I moved that logic into `authentication.py` (where the `UserInfo` object is created). The benefit of this change is that all authentication decisions are now centralized in `authentication.py` and we treat pets and their users equally throughout Bond which is how pets should be treated in Terra.

A user's enabled status is cached for 10 minutes now. While it would be nice to be able to instantly disable a user, this is still far better than before when we weren't checking enabled status at all. ~One thing to be conscious of is we are caching the user's enabled status for an hour which means the user would not be fully disabled in Bond until the cached user info from Sam expires. I think the best balance between disabling users quickly and not overwhelming Sam with user info requests is to shorten the cache expiration from an hour down to \~10 minutes. It's also worth noting that previously we didn't check whether the user was enabled in Sam *at all*. An hour is definitely better than never, but we should consider reducing the time further.~

In all cases:


tests/datastore_emulator/fence_token_vending_test.py
```diff
@@ -1,73 +0,0 @@
-from bond_app.fence_token_vending import FenceTokenVendingMachine
```
marctalbott, line 1: I initially wrote this test to make sure that we were clearing out the SA key that we were caching when getting a SA key for a user. We no longer cache this SA key, so I deleted the test.

bond_app/authentication.py
```diff
@@ -74,17 +77,21 @@ def require_user_info(self, request_state, token_info_fn=_token_info):
...
-            expires_in = min([user_info.expires_in, self.config.max_token_life])
+            expires_in = min([google_user_info.expires_in, self.config.max_token_life])
             logging.debug("caching token %s for %s seconds", token, expires_in)
-            self.cache_api.add(key='access_token:' + token, value=user_info, expires_in=expires_in)
+            self.cache_api.add(key='access_token:' + token, value=google_user_info, expires_in=expires_in)
         else:
             logging.debug("auth token cache hit for token %s", token)
 
-        return user_info
+        sam_user_info = self.sam_api.user_info(google_user_info)
```
dvoet: from a code organization perspective I would prefer that the caching code be here and leave the sam_api as just the raw sam api calls. sam_api should have limited logic, it should just do the sam things when asked (in other words, when I call sam_api I want to actually call sam). Here is a higher level bit of code that has more business logic and can be more sophisticated.

dvoet (changes requested):

bond_app/authentication.py
```diff
@@ -74,17 +77,21 @@ def require_user_info(self, request_state, token_info_fn=_token_info):
...
-        user_info = self.cache_api.get(key='access_token:' + token)
-        if user_info is None:
-            user_info = self._fetch_user_info(token, token_info_fn)
+        google_user_info = self.cache_api.get(key='access_token:' + token)
+        if google_user_info is None:
+            google_user_info = self._fetch_user_info(token, token_info_fn)
             # cache for 10 minutes or until token expires
-            expires_in = min([user_info.expires_in, self.config.max_token_life])
+            expires_in = min([google_user_info.expires_in, self.config.max_token_life])
             logging.debug("caching token %s for %s seconds", token, expires_in)
```
gpolumbo-broad, line 85: How would you feel about changing some of these `debug` log statements to `info`?  I kind of feel like one of the biggest struggles we have with Bond is how much detail we get out of logs, so we should turn the volume up where appropriate, and I think tracking when we get cache hits vs. when we are calling out to other services is everyday stuff that could be helpful in the logs.

bond_app/authentication.py
```diff
@@ -74,17 +77,21 @@ def require_user_info(self, request_state, token_info_fn=_token_info):
...
-            self.cache_api.add(key='access_token:' + token, value=user_info, expires_in=expires_in)
+            self.cache_api.add(key='access_token:' + token, value=google_user_info, expires_in=expires_in)
         else:
             logging.debug("auth token cache hit for token %s", token)
 
-        return user_info
+        sam_user_info = self.sam_api.user_info(google_user_info)
+        if sam_user_info is None or not sam_user_info[SamKeys.USER_ENABLED_KEY]:
```
gpolumbo-broad, line 99: Be careful of the truthiness of this statement:  https://stackoverflow.com/questions/18491777/truth-value-of-a-string-in-python

I do not know if this is problematic or not as it stands, depends on how the cache api and the `requests` library are populating the field for this object.

bond_app/authentication.py
```diff
@@ -74,17 +77,21 @@ def require_user_info(self, request_state, token_info_fn=_token_info):
...
+            raise exceptions.Unauthorized("user not found in sam")
```
gpolumbo-broad: I can appreciate having this be the error raised, in the spirit of not returning a message that indicates the user exists but is just not enabled.

That said, we may benefit from a log statement here to show the response from Sam, might be helpful in debugging?

bond_app/bond.py
```diff
@@ -39,83 +37,70 @@ def build_authz_url(self, scopes, redirect_uri, state=None):
...
-        provider.  If a refresh token cannot be found for the user_id provided, a NotFound will be raised.
-        :param user_info: Information of the user who issued the request to Bond (not necessarily the same as
-        the username for whom the refresh token was issued by the OAuth provider)
+        provider.  If a refresh token cannot be found for the sam_user_id provided, a NotFound will be raised.
+        :param sam_user_id: Id stored in Sam for user who initiated request
         :return: Two values: An Access Token string, datetime when that token expires
         """
-        user_id = self._fetch_user_id(user_info)
-        refresh_token = self.refresh_token_store.lookup(user_id, self.provider_name)
+        refresh_token = self.refresh_token_store.lookup(sam_user_id, self.provider_name)
```
gpolumbo-broad, line 70: Wherever we lookup the Refresh Token for users, we might have a problem after we launch this.  Because I think pretty much everything calls Sam using a pet, which means the refresh tokens currently in the system for linked users are under the pet's key.  Now we're switching to the User's key, which is the right thing to do and what we want, but I think it means that all existing links won't work anymore because we won't be able to find the refresh token for existing links.

bond_app/authentication.py
```diff
@@ -48,21 +49,23 @@ def __init__(self, accepted_audience_prefixes, accepted_email_suffixes, max_toke
...
+    def __init__(self, config, cache_api, sam_api):
         """
         :param config: An AuthenticationConfig instance.
         :param cache_api: A CacheApi instance.
         """
         self.config = config
         self.cache_api = cache_api
+        self.sam_api = sam_api
 
     def require_user_info(self, request_state, token_info_fn=_token_info):
```
gpolumbo-broad: Should we rename this?  `authn_authz_user`, `authorize_user`, `verify_access_token`, `validate_user`, ... ?

gpolumbo-broad (changes requested):

bond_app/authentication.py
marctalbott: I was trying to think of a good name because I don't think `require_user_info` is quite right anymore, but didn't come up with anything so just left it as is. How about `authenticate_user`? We aren't really doing any authz imo, we're just translating the token into the Sam user id

bond_app/bond.py
marctalbott: `self._fetch_user_id(user_info)` connects the pet to the user, so we're storing refresh tokens under the user. We do cache SA keys for the pet instead of the user in the FenceTokenVendingMachine, so we might need to clear those out, but they'll also expire eventually so it is a problem that would go away on its own

bond_app/authentication.py
marctalbott: Yeah, I wasn't quite sure what to do with this error message either. I left it in the interest of not returning too informative of a message, but I kind of want to change it to something along the lines of "user is not registered in Sam" or even just removing the message entirely and just returning the 401 default message

marctalbott: With added logging though, so _we_ can at least see what's happening

bond_app/authentication.py
marctalbott: I'm worried that we'd flood the logs if these were `info` level. As I understand it, we want this caching so that when users are running workflows on potentially many VMs, we don't slam Google or Sam trying to get user info. If these were all `info` statements, anytime a user ran a workflow it could just show flood the logs

bond_app/authentication.py
marctalbott: I was worried about this too. I tried to test it in the python console, but I'm not 100% sure my test was sufficient. This is what I did

> `json.loads("{\"userSubjectId\": \"1234\", \"enabled\": true}")`
> `{'userSubjectId': '1234', 'enabled': True}`

This makes me believe that `json.loads` will recognize it as a boolean so that we can treat it as such later on when accessing the dict, but I can try to do more testing to be sure

bond_app/authentication.py
gpolumbo-broad: Fine with me.   I would say we are doing the bare minimum authz in that we are checking to see that the user is enabled in Sam, so maybe we fudge it and go with `auth_user` and the reader can choose-their-own-adventure as to whether they think of it as `authz` or `authn` in their own brains?  🤣

bond_app/bond.py
gpolumbo-broad: OK, cool.  Just wanted to double check whether we'd need to alert users as part of this release.  Sounds like we're OK.

bond_app/authentication.py
gpolumbo-broad: Right, in terms of the message returned, I would want to return the most descriptive and precise message we can that is reasonable and secure.  Maybe sec ops has an opinion about whether we can say "user not enabled in Sam".  

At the very least, get that information into the logs so that when a user does have a problem and they go, "But I'm registered!" we can look in the logs and get the details.

bond_app/authentication.py
gpolumbo-broad: Right, my worry would be around:

`foo = json.loads("{\"userSubjectId\": \"1234\", \"enabled\": \"false\"}")`

because I think the truthiness rules would evaluate `foo.enabled` as `True` in this case.

bond_app/authentication.py
gpolumbo-broad: well, I think app engine is logging entries per request anyway, so no matter what, when a workflow is run, if it's banging on Bond, it's going to log (probably multiple) entries per request anyway.  I don't know how many log entries app engine is making for each request right now, but even if it's 2 or 3 entries per request now, I don't see harm in going to 5 or 6 or even 10 so long as the information logged is useful.  I think this is a big reason we've struggled in debugging Bond stuff in the past is just a lack of logging.

bond_app/authentication.py
marctalbott: I did some extra testing and it does work like we want it too, so this should be fine, but good call on double checking it.

> result = requests.get(url=base_url + '/register/user/v2/self/info', headers=headers)
> sam_user_info = json.loads(result.content)
> sam_user_info
> {'enabled': True, 'userEmail': 'marc.testbott@gmail.com', 'userSubjectId': '106960538915967416130'}

and an if statement as a sanity check
> if sam_user_info["enabled"]:
...     print('user is enabled')
... 
user is enabled

bond_app/authentication.py
marctalbott: How do you like the message I changed it to, "could not authenticate with Sam" along with a log message that includes the user's email, google subject id, and the response from Sam?

gpolumbo-broad (approved):

bond_app/authentication.py
dvoet: this looks like a debug level message to me. We should not be inflating the message priority to get more verbosity out of our code, we should instead manipulate the logging level reported. If we want to see debug level statements in stack driver we should make sure debug log messages are emitted.

dvoet (approved):

PR #156, opened 2020-10-01
[CA-1038] More tolerant of expired refresh tokens when unlinking

marctalbott:
Ticket: [CA-1038](https://broadworkbench.atlassian.net/browse/CA-1038)

In all cases:

### DIFF
--- a/bond_app/fence_api.py
+++ b/bond_app/fence_api.py
@@ -33,7 +33,7 @@ class FenceApi:
         headers = {'Authorization': 'Bearer ' + access_token}
         result = requests.delete(url=self.delete_service_account_url + key_id, headers=headers)
         logging.info("request: DELETE {} - status code: {}".format(self.delete_service_account_url, result.status_code))
-        if result.status_code // 100 != 2 or result.status_code // 100 != 4:
+        if result.status_code // 100 != 2 and result.status_code // 100 != 4:
             raise exceptions.InternalServerError("fence status code {}, error body {}".format(result.status_code, result.content))
 
     def revoke_refresh_token(self, refresh_token):
--- a/bond_app/fence_api.py
+++ b/bond_app/fence_api.py
@@ -1,5 +1,6 @@
 from werkzeug import exceptions
 import requests
+import logging
 
 
 class FenceApi:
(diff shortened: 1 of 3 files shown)


bond_app/fence_api.py
```diff
@@ -31,7 +32,8 @@ def delete_credentials_google(self, access_token, key_id):
         """
         headers = {'Authorization': 'Bearer ' + access_token}
         result = requests.delete(url=self.delete_service_account_url + key_id, headers=headers)
-        if result.status_code // 100 != 2:
+        logging.info("request: DELETE {} - status code: {}".format(self.delete_service_account_url, result.status_code))
+        if result.status_code // 100 != 2 or result.status_code // 100 != 4:
```
dvoet: this should be `and` not `or`

dvoet (changes requested):

bond_app/fence_api.py
```diff
@@ -31,7 +32,8 @@ def delete_credentials_google(self, access_token, key_id):
         """
         headers = {'Authorization': 'Bearer ' + access_token}
         result = requests.delete(url=self.delete_service_account_url + key_id, headers=headers)
-        if result.status_code // 100 != 2:
+        logging.info("request: DELETE {} - status code: {}".format(self.delete_service_account_url, result.status_code))
+        if result.status_code // 100 != 2 and result.status_code // 100 != 4:
```
gpolumbo-broad, line 39: Can we leave a comment here on why we are accepting `4xx` as OK.

gpolumbo-broad (changes requested):

gpolumbo-broad:
And I'm still an advocate of directly deleting the item from the cache when we know we can remove it at deletion time rather than waiting for the expiry logic to purge it, but I don't have a good technical or logical reason why need to do it, so it's fine.  To me, it's sort of like dropping a piece of trash on the street when I see the street sweeper coming.  It's going to get swept up, so I can just leave it, but I'm probably going to pick it up and throw it in the trash. :)

marctalbott:
> And I'm still an advocate of directly deleting the item from the cache when we know we can remove it at deletion time rather than waiting for the expiry logic to purge it, but I don't have a good technical or logical reason why need to do it, so it's fine. To me, it's sort of like dropping a piece of trash on the street when I see the street sweeper coming. It's going to get swept up, so I can just leave it, but I'm probably going to pick it up and throw it in the trash. :)

We were directly deleting SA keys that had been associated with user ids from the cache. We don't do that anymore after #155 because of the discrepancy it was causing between pets and users calling Bond. That's why this PR removes that direct deletion, since there's no longer anything to delete

gpolumbo-broad (approved):

dvoet (approved):

PR #173, opened 2021-05-07
CA-1341 dpath should ALLOW_EMPTY_STRING_KEYS

gpolumbo-broad:
Updating the dpath lib and adding dpath option to ALLOW_EMPTY_STRING_KEYS

In all cases:

### DIFF
--- a/bond_app/jwt_token.py
+++ b/bond_app/jwt_token.py
@@ -1,6 +1,7 @@
 import jwt
 from datetime import datetime
 import dpath.util
+import dpath.options
 
 
 class JwtToken:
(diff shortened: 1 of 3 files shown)


bond_app/jwt_token.py
```diff
@@ -11,5 +12,7 @@ def __init__(self, encoded_str, user_name_path_expr):
         :param user_name_path_expr: forward slash delimited path to user name within JWT
         """
         self.raw_dict = jwt.decode(encoded_str, verify=False)
+        # See https://broadworkbench.atlassian.net/browse/CA-1341, needed for dpath to work when dict has "" keys
```
s-rubenstein, line 15: (As a note, this is unassigned right now, so should probably be assigned to you?)

requirements.txt
```diff
@@ -4,7 +4,7 @@ requests-oauthlib==1.3.0
 requests-toolbelt==0.8.0
 pyjwt==1.6.4
 mock==2.0.0
-dpath==1.4.2
+dpath==2.0.1
```
s-rubenstein, line 7: Do we have testing to ensure nothing breaks with this upgrade? I see the test below, but since that test is changing, I wanted to check if this was this used anywhere else in code that we need to be aware of.

marctalbott (approved):

requirements.txt
gpolumbo-broad: correct, only place it is used and it is tested.

s-rubenstein (approved):
LGTM

PR #175, opened 2021-05-27
CA-1205 Bond release checklist template

gpolumbo-broad:
Instead of using github issues, let's try to do everything in Jira.

This solution uses the Issue Checklist Plugin for Jira, https://herocoders.atlassian.net/wiki/spaces/IC/overview

The new .md file contains the source code that can be used in Bond Release Checklist template in Jira.  The checklist was removed from the README so that there is only 1 source of truth for what the release checklist is.

Users should never modify the checklist in Jira directly, instead, they should modify the release-checklist.md in this repo first, get it reviewed as normal, and only after it has merged should that files contents be copy/pasted into the Jira checklist template.  We don't have a way to enforce this automatically so it must be enforced by policy.


marctalbott (approved):
Have you run this by Gary? I like the idea of doing it all in Jira, but we should make sure that it satisfies the compliance need that was the original motivation for the issue release checklist

gdlex4015 (commented):
Just so I understand, you are going to keep a template in github, but when doing releases the actual ticket is going to be created in Jira? I am fine with that, as I think eventually all of the services will go that way.

gdlex4015 (approved):

PR #178, opened 2022-04-13
Mspector bt-634 m1 mac instructions

mspector:
Update the README to include a short note about setup on an M1 Mac. The more consequential change is an upgrade from Python `3.7.4` --> `3.7.13`, and pinning the version of certain dependencies. This should be a non-breaking change, but may have deployment implications.

In all cases:

- [x] Get two thumbsworth of review and PO signoff if necessary. 


andy7i (approved):
LGTM!

requirements.txt
```diff
@@ -21,3 +21,5 @@ gunicorn==20.0.4
 flask-cors==3.0.10
 flask-swagger-ui==3.20.9
 PyYaml==5.4
+jinja2<3.1.0
```
jgainerdewar: Curious why this is `<` rather than `==` or `>`.

mspector: It's `<` and not `>` because there was a breaking API change after this version.

As for why it's not `==`... it might as well be! I'll change that.

jgainerdewar (approved):

requirements.txt
```diff
@@ -21,3 +21,5 @@ gunicorn==20.0.4
...
+itsdangerous==2.0.1
```
breilly2, line 25: Can we not use something newer? Current is `2.1.2`.

mspector: We probably can, but there are breaking API changes after this version, so we’d need to make some minor updates elsewhere.

That makes sense to do, in my opinion, if we want to view this as a renovation more than a repair job.

breilly2 (approved):
👍  I didn't test, but looks reasonable to me.

mspector:
@andy7i , should we be concerned about the failing circleci test?

andy7i:
> @andy7i , should we be concerned about the failing circleci test?

I don't think it's your PR. looks like this is affecting all PRs right now. https://github.com/DataBiosphere/bond/pull/180
I'll look into it but I might not get to it today.

andy7i:
@mspector This should fix the issue. https://github.com/DataBiosphere/bond/pull/181 -- feel free to merge it into your branch!

PR #179, opened 2022-04-15
[BT-622] Cache access tokens

mspector:
Caches access tokens using the existing Datastore cache API. Builds upon https://github.com/DataBiosphere/bond/pull/178

In all cases:

### DIFF
--- a/bond_app/bond.py
+++ b/bond_app/bond.py
@@ -106,16 +106,18 @@ class Bond:
                 expires_at = cached_access_entry.expires_at
                 access_token = cached_access_entry.value
                 expires_in = (expires_at - datetime.now()).total_seconds()
-                logging.info(
-                    f"Retrieved access token from cache. "
-                    f"Cache will be refreshed in {expires_in - refresh_threshold:.2f} seconds."
+                logging.debug(
+                    "Retrieved access token from cache. " +
+                    f"Access token will expire in {expires_in - refresh_threshold:.2f} seconds. " +
+                    f"SAM user ID: {sam_user_id}. Provider: {self.provider_name}"
                 )
             else:
                 access_token, expires_at = self.generate_access_token(sam_user_id, refresh_token=refresh_token)
                 expires_in = (expires_at - datetime.now()).total_seconds()
-                logging.info(
-                    f"Generated new access token. "
-                    f"Cache will be refreshed in {expires_in - refresh_threshold:.2f} seconds."
+                logging.debug(
+                    "Generated new access token. " +
+                    f"Access token will expire in {expires_in - refresh_threshold:.2f} seconds. " +
+                    f"SAM user ID: {sam_user_id}. Provider: {self.provider_name}"
                 )
                 
                 self.cache_api.add(
--- a/bond_app/bond.py
+++ b/bond_app/bond.py
@@ -98,7 +98,7 @@ class Bond:
         """
         refresh_token = self.refresh_token_store.lookup(sam_user_id, self.provider_name)
         if refresh_token is not None:
-            cached_access_data = self.cache_api.get(namespace="AccessTokens", key=sam_user_id)
+            cached_access_data = self.cache_api.get(namespace=f"{self.provider_name}:AccessTokens", key=sam_user_id)
             if cached_access_data and cached_access_data.get(FenceKeys.ACCESS_TOKEN):
                 expires_at = cached_access_data.get(FenceKeys.EXPIRES_AT)
                 access_token = cached_access_data.get(FenceKeys.ACCESS_TOKEN)
@@ -118,7 +118,7 @@ class Bond:
                 )
                 
                 self.cache_api.add(
-                    namespace="AccessTokens", 
+                    namespace=f"{self.provider_name}:AccessTokens", 
                     key=sam_user_id, 
                     value={
                         FenceKeys.EXPIRES_AT: expires_at,
--- a/bond_app/bond.py
+++ b/bond_app/bond.py
@@ -98,33 +98,33 @@ class Bond:
         """
         refresh_token = self.refresh_token_store.lookup(sam_user_id, self.provider_name)
         if refresh_token is not None:
-            cached_access_entry = self.cache_api.get_entry(namespace="AccessTokens", key=sam_user_id)
-            if (
-                cached_access_entry and 
-                cached_access_entry.expires_at - datetime.now() > timedelta(seconds=refresh_threshold)
-            ):
-                expires_at = cached_access_entry.expires_at
-                access_token = cached_access_entry.value
-                expires_in = (expires_at - datetime.now()).total_seconds()
+            cached_access_data = self.cache_api.get(namespace="AccessTokens", key=sam_user_id)
+            if cached_access_data and cached_access_data.get(FenceKeys.ACCESS_TOKEN):
+                expires_at = cached_access_data.get(FenceKeys.EXPIRES_AT)
+                access_token = cached_access_data.get(FenceKeys.ACCESS_TOKEN)
+                expires_in_cache = (expires_at - datetime.now()).total_seconds() - refresh_threshold
                 logging.debug(
                     "Retrieved access token from cache. " +
-                    f"Access token will expire in {expires_in - refresh_threshold:.2f} seconds. " +
+                    f"Access token will expire in {expires_in_cache:.2f} seconds. " +
                     f"SAM user ID: {sam_user_id}. Provider: {self.provider_name}"
                 )
             else:
                 access_token, expires_at = self.generate_access_token(sam_user_id, refresh_token=refresh_token)
-                expires_in = (expires_at - datetime.now()).total_seconds()
+                expires_in_cache = (expires_at - datetime.now()).total_seconds() - refresh_threshold
                 logging.debug(
                     "Generated new access token. " +
-                    f"Access token will expire in {expires_in - refresh_threshold:.2f} seconds. " +
+                    f"Access token will expire in {expires_in_cache:.2f} seconds. " +
                     f"SAM user ID: {sam_user_id}. Provider: {self.provider_name}"
                 )
                 
                 self.cache_api.add(
                     namespace="AccessTokens", 
                     key=sam_user_id, 
-                    value=access_token,
-                    expires_in=expires_in,
+                    value={
+                        FenceKeys.EXPIRES_AT: expires_at,
+                        FenceKeys.ACCESS_TOKEN: access_token,
+                    },
+                    expires_in=expires_in_cache,
                 )
             return access_token, expires_at
         else:
--- a/bond_app/bond.py
+++ b/bond_app/bond.py
@@ -3,7 +3,7 @@ import logging
 from .jwt_token import JwtToken
 
 from werkzeug import exceptions
-
+from dataclasses import dataclass
 
 class Bond:
     def __init__(self,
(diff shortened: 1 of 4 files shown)


bond_app/bond.py
```diff
@@ -60,23 +60,61 @@ def exchange_authz_code(self, authz_code, redirect_uri, sam_user_id):
                                       jwt_token.username, self.provider_name)
         return jwt_token.issued_at, jwt_token.username
 
-    def generate_access_token(self, sam_user_id):
+
+    def generate_access_token(self, sam_user_id, refresh_token=None):
         """
-        Given a user, lookup their refresh token and use it to generate a new refresh token from their OAuth
+        Given a user, lookup their refresh token (if not provided) and use it to generate a new refresh token (<-should this be "access token"?) from their OAuth
```
mspector: see inline "comment" -- given my understanding, I think this should read as follows:
> Given a user, lookup their refresh token (if not provided) and use it to generate a new *access token* from their OAuth provider.

bond_app/bond.py
```diff
@@ -60,23 +60,61 @@ def exchange_authz_code(self, authz_code, redirect_uri, sam_user_id):
...
+            if (
+                cached_access_entry and 
+                cached_access_entry.expires_at - datetime.now() > timedelta(seconds=refresh_threshold)
+            ):
+                expires_at = cached_access_entry.expires_at
+                access_token = cached_access_entry.value
+                expires_in = (expires_at - datetime.now()).total_seconds()
+                logging.info(
+                    f"Retrieved access token from cache. "
+                    f"Cache will be refreshed in {expires_in - refresh_threshold:.2f} seconds."
```
mspector: I found logging this helpful, but do you all want it here?

bond_app/datastore_cache_api.py
```diff
@@ -29,11 +29,15 @@ def add(self, key, value, expires_in=0, namespace=None):
                    expires_at=DatastoreCacheApi._calculate_expiration(expires_in)).put()
         return True
 
-    def get(self, key, namespace=None):
+    def get_entry(self, key, namespace=None):
         entry = DatastoreCacheApi._build_cache_key(key, namespace).get()
         if not entry or entry.expires_at < datetime.datetime.now():
             return None
-        return entry.value
+        return entry
```
mspector: I went back and forth on whether or not I should be sending CacheEntry objects to other parts of the code, but these modules are on the same "level", so it seemed ok, and it means we don't have to package up this data in the cache in some other way (a dict, a tuple, etc.)

bond_app/bond.py
```diff
@@ -60,23 +60,61 @@ def exchange_authz_code(self, authz_code, redirect_uri, sam_user_id):
...
         else:
             raise exceptions.NotFound(
                 "Could not find refresh token for sam_user_id: {} provider_name: {}\nConsider relinking your account to Bond.".format(
                     sam_user_id, self.provider_name))
 
+
+    def get_access_token(self, sam_user_id, refresh_token=None, refresh_threshold: int = 600):
+        refresh_token = refresh_token or self.refresh_token_store.lookup(sam_user_id, self.provider_name)
+        if refresh_token is not None:
+            cached_access_entry = self.fence_tvm.cache_api.get_entry(namespace="AccessTokens", key=sam_user_id)
```
mspector: It seemed reasonable at the time, but now I'm not sure that the `FenceTokenVendingMachine` is the right place to access this cache API. Let me know what you think...

bond_app/bond.py
jgainerdewar: Yeah, this seems like an error to me.

bond_app/bond.py
```diff
@@ -60,23 +60,61 @@ def exchange_authz_code(self, authz_code, redirect_uri, sam_user_id):
...
             expires_at = datetime.fromtimestamp(token_response.get(FenceKeys.EXPIRES_AT))
+
             return token_response.get("access_token"), expires_at
         else:
             raise exceptions.NotFound(
                 "Could not find refresh token for sam_user_id: {} provider_name: {}\nConsider relinking your account to Bond.".format(
                     sam_user_id, self.provider_name))
 
+
+    def get_access_token(self, sam_user_id, refresh_token=None, refresh_threshold: int = 600):
```
jgainerdewar: Can you add a docstring for this method?

bond_app/bond.py
jgainerdewar: I'd be interested in hearing from someone who knows this codebase and its prod operations better (is prod configured to log INFO?) , but naively I'd suggest downgrading to DEBUG. 

Especially if we do make this visible in prod, may want to change to a single line and include the user id (and provider name?) for easier log exploration.

bond_app/bond.py
jgainerdewar: Honest question and not (yet?) veiled suggestion: Did you consider adding a structure like the `[Refresh]TokenStore` to hold access tokens? 

I'm having trouble tracking the various types of tokens and places we can stick data, but I'm not sure it feels natural to put them here. It looks like the cache api on `FenceTokenVendingMachine` wasn't actually used before this change, which makes me suspicious.

jgainerdewar (commented):
Looks good overall, left a few comments. Once we're confident in the general shape of this, let's add some unit tests!

bond_app/bond.py
mspector: I didn't consider that, but it makes sense. The `TokenStore` class seems to have a lot in common with `DatastoreCacheApi`, almost like it evolved in parallel, or one represents a newer vision of how tokens will be cached in the future.

> It looks like the cache api on FenceTokenVendingMachine wasn't actually used before this change, which makes me suspicious.

100% agree with this.

It's probably simplest to create a parallel `AccessTokenStore` class, but we may want someone more familiar with a higher level view of this repository to chime in about the direction we should go.

bond_app/bond.py
mspector: This is more of a "fun fact" comment than a rebuttal of yours, but FWIW: this actually is one line :)

It's just a particularly opaque way of doing string concatenation. I should change it. Let's see what someone more familiar with Bond has to say about your more important questions, and I'll make it clearer in the next update.

bond_app/bond.py
dvoet: agreed

bond_app/bond.py
dvoet: I would not use `self.fence_tvm.cache_api` but rather pass in a `cache_api` instance to this class via constructor

bond_app/bond.py
dvoet: definitely debug

bond_app/bond.py
dvoet: why is `refresh_token` a parameter?

bond_app/bond.py
```diff
@@ -60,23 +60,61 @@ def exchange_authz_code(self, authz_code, redirect_uri, sam_user_id):
...
+                logging.info(
+                    f"Retrieved access token from cache. "
+                    f"Cache will be refreshed in {expires_in - refresh_threshold:.2f} seconds."
+                )
+            else:
+                access_token, expires_at = self.generate_access_token(sam_user_id, refresh_token=refresh_token)
+                expires_in = (expires_at - datetime.now()).total_seconds()
+                logging.info(
+                    f"Generated new access token. "
+                    f"Cache will be refreshed in {expires_in - refresh_threshold:.2f} seconds."
```
dvoet: the terminology `Cache will be refreshed in` is confusing because it sounds like the entire cache. How about something like `Access token will expire in`

bond_app/bond.py
```diff
@@ -60,23 +60,61 @@ def exchange_authz_code(self, authz_code, redirect_uri, sam_user_id):
...
+                )
+                
+                self.fence_tvm.cache_api.add(
+                    namespace="AccessTokens", 
+                    key=sam_user_id, 
+                    value=access_token,
+                    expires_in=expires_in,
```
dvoet: `expires_in` is when the item will be removed from the cache. This value should _not_ be when the token itself expires but should have the `refresh_threshold` already subtracted. The cache itself should contain a structure that has both the access token and its real expiration. Then undo your `datastore_cache_api` changes. Also the date logic on line 89 should go away. If the item is in the cache, use it.

dvoet (changes requested):

bond_app/bond.py
mspector: good catch -- I see that it's not needed here. This was probably bad copy/paste thinking at work.

bond_app/bond.py
mspector: thanks @dvoet, that sounds good to me.

I am a bit curious about `TokenStore` vs `DatastoreCacheApi`. But if that's out-of-scope, it's all good with me.

bond_app/bond.py
mspector: good call. I'll change it to your suggestion

bond_app/bond.py
```diff
@@ -60,23 +62,76 @@ def exchange_authz_code(self, authz_code, redirect_uri, sam_user_id):
...
+        and that token has greater than `refresh_threshold` seconds before expiration, return that token. 
+        Otherwise, generate a new token.
+        
+        If a refresh token cannot be found for the sam_user_id provided, a NotFound will be raised.
+        :param sam_user_id: Id stored in Sam for user who initiated request
+        :return: Two values: An Access Token string, datetime when that token expires
+        """
+        refresh_token = self.refresh_token_store.lookup(sam_user_id, self.provider_name)
+        if refresh_token is not None:
+            cached_access_data = self.cache_api.get(namespace="AccessTokens", key=sam_user_id)
```
dvoet: `namespace` needs to be provider specific. Each provider has its own set of access tokens.

bond_app/bond.py
```diff
@@ -60,23 +62,76 @@ def exchange_authz_code(self, authz_code, redirect_uri, sam_user_id):
...
+            if cached_access_data and cached_access_data.get(FenceKeys.ACCESS_TOKEN):
+                expires_at = cached_access_data.get(FenceKeys.EXPIRES_AT)
+                access_token = cached_access_data.get(FenceKeys.ACCESS_TOKEN)
+                expires_in_cache = (expires_at - datetime.now()).total_seconds() - refresh_threshold
```
dvoet: can you just log `expires_at` instead of calculating the `expires_in_cache`? Doing the calculation here seems odd.

dvoet (changes requested):

bond_app/bond.py
```diff
@@ -60,23 +62,76 @@ def exchange_authz_code(self, authz_code, redirect_uri, sam_user_id):
...
+                    f"Access token will expire in {expires_in_cache:.2f} seconds. " +
+                    f"SAM user ID: {sam_user_id}. Provider: {self.provider_name}"
+                )
+                
+                self.cache_api.add(
+                    namespace="AccessTokens", 
+                    key=sam_user_id, 
+                    value={
+                        FenceKeys.EXPIRES_AT: expires_at,
+                        FenceKeys.ACCESS_TOKEN: access_token,
```
dvoet: `FenceKeys` is for a different use case, I would not reuse them here. I would create a class to hold a Fence access token instead. It reads better that dealing with a dict.

mspector: would you put the Fence access token class here in `bond.py`? or elsewhere?

dvoet (approved):

bond_app/bond.py
```diff
@@ -77,6 +83,52 @@ def generate_access_token(self, sam_user_id):
                 "Could not find refresh token for sam_user_id: {} provider_name: {}\nConsider relinking your account to Bond.".format(
                     sam_user_id, self.provider_name))
 
+
+    def get_access_token(self, sam_user_id, refresh_threshold: int = 600):
+        """
+        Given a user, lookup their refresh token (if not provided) and use it to retrieve an access token from their OAuth
```
breilly2:
```suggestion
        Given a user, lookup their refresh token and use it to retrieve an access token from their OAuth
```
No `refresh_token` parameter to this function.

tests/unit/bond_test.py
```diff
@@ -114,6 +116,117 @@ def test_generate_access_token(self):
...
+                    {})
+        
+        token = str(uuid.uuid4())
+        self.refresh_token_store.save(user_id=self.user_id,
+                                      refresh_token_str=token,
+                                      issued_at=datetime.fromtimestamp(self.issued_at_epoch),
+                                      username=self.name,
+                                      provider_name=provider_name)
+
+        # verify that the new `expired_at` value is returned with `get_access_token`,
```
breilly2: Nit:
```suggestion
        # verify that the new `expires_at` value is returned with `get_access_token`,
```
Also on lines 212 and 223.

tests/unit/bond_test.py
```diff
@@ -114,6 +116,117 @@ def test_generate_access_token(self):
...
+
+        # verify that the new `expires_at` value is returned with `generate_access_token`.
+        access_token, expires_at = bond.generate_access_token(self.user_id)
+        self.assertEqual(self.fake_access_token, access_token)
+        self.assertEqual(datetime.fromtimestamp(expires_at_new), expires_at)
+
+        # verify that the cached `expires_at` value is returned with `get_access_token`
+        access_token, expires_at = bond.get_access_token(self.user_id)
+        self.assertEqual(self.fake_access_token, access_token)
+        self.assertEqual(datetime.fromtimestamp(expires_at_initial), expires_at)
```
breilly2: This looks like it's checking the same thing as before the call to `generate_access_token`. Are you verifying that `generate_access_token` does not affect the cache?

TOL: I think that might be an overly specific test. If you do keep it, I think it would be helpful to be explicit about the behavior you're intending to verify.

breilly2 (approved):
👍 Just minor nits. Otherwise looks good!

tests/unit/bond_test.py
mspector: I see your point that this may be overly specific. Ironically, I put it there to make it clearer to the reader what the "real" test (the call to `get_access_token`) is doing, by providing a point of contrast. But it seems like I've done the opposite.

I think it's worth having a quick conversation about this

jgainerdewar (approved):
LGTM pending the outcome of the conversation with Brian about the test.

PR #184, opened 2022-05-16
[TOAZ-132] Update bond to support B2C authentication

rtitle:
https://broadworkbench.atlassian.net/browse/TOAZ-132

Updated bond to support authenticating with B2C tokens, with a fallback to Google tokeninfo. (Most services use the Apache proxy for this, but bond does it in python code. Bond is the last remaining service to be updated 😅 .)

I'm not extremely familiar with bond, but it looks like the user's oauth token is only used to authorize with Sam, and downstream code uses the Sam user id. So I think the scope of these changes should be ok.

Use `pyjwt` library to verify the JWT signature using the jwks endpoint. Also validate that all the claims are valid (sub, aud, iat, exp, email).

Testing performed:
* Added unit tests for JWT authentication
* Existing automation tests passed (with google tokens)
   * It would be nice to add automation tests that use B2C tokens, but we don't have great support for that now so I think it's out of scope of this PR. 
* Deployed to a fiab, tested with a local B2C-enabled Terra UI pointing to the fiab. The calls to bond authenticate successfully but return 404:
   ```
   {"error":{"code":404,"errors":[{"domain":"global","message":"fence link does not exist. Consider re-linking your account.","reason":"notFound"}],"message":"fence link does not exist. Consider re-linking your account."}}
   ```
   This happens for both Google and B2C so I _think_ it's expected? **Q: is there is a better way to test bond with a local Terra UI?**

----

In all cases:

### DIFF
--- a/bond_app/authentication.py
+++ b/bond_app/authentication.py
@@ -111,8 +111,8 @@ class Authentication:
             if self.config.b2c_authority_endpoint:
                 try:
                     user_info = self._fetch_user_info_from_jwt(token)
-                except:
-                    logging.info("Failed to parse token as a JWT. Falling back to google tokeninfo...")
+                except Exception as e:
+                    logging.warning('Failed to parse token as a JWT: {}. Falling back to google tokeninfo...'.format(e))
             
             # Fall back to Google tokeninfo if that fails.
             if user_info is None:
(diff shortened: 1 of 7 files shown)


bond_app/authentication.py
```diff
@@ -40,13 +42,44 @@ def _token_info(token):
 
     return result.content
 
+def _decode_jwt(token, jwks_uri, audience):
```
rtitle, line 45: Used https://pyjwt.readthedocs.io/en/stable/usage.html#retrieve-rsa-signing-keys-from-a-jwks-endpoint as a guide for how to validate JWTs using `pyjwt`.

tests/unit/authentication_test.py
```diff
@@ -2,27 +2,63 @@
...
+ALGORITHM = "RS256"
+PUBLIC_KEY_ID = "sample-key-id"
 
 class TestRequestState:
     def __init__(self, auth_header):
         self.headers = {}
         if auth_header is not None:
             self.headers['Authorization'] = auth_header
 
+class MockOidcServer:
```
rtitle, line 25: Creates a mock `jwks` endpoint to be used for encoding and decoding valid JWTs for testing.

Used https://dev.to/filipemir/mocking-auth0-tokens-in-python-and-beyond-5dki as a guide. Was kinda fun to put together, had to dust off some python skills.

rtitle:
Existing automation tests passed. It would be nice to have some automation test cases that use a B2C token. But I'm thinking that's out of scope of this PR..

bond_app/jwt_token.py
```diff
@@ -11,7 +11,7 @@ def __init__(self, encoded_str, user_name_path_expr):
         :param encoded_str: encoded JWT
         :param user_name_path_expr: forward slash delimited path to user name within JWT
         """
-        self.raw_dict = jwt.decode(encoded_str, verify=False)
+        self.raw_dict = jwt.decode(encoded_str, options={'verify_signature': False})
```
rtitle, line 14: API change due to to `pyjwt` upgrade

dvoet:
I did not review this but I think there could be a better approach. Ditch the existing authentication code and call the [sam user info api](https://sam.dsde-prod.broadinstitute.org/#/Users/getUserStatusInfo) instead. And while you are at it, check the `enabled` field of the response.

rtitle:
Gonna close this for now and try to the simpler approach above

PR #185, opened 2022-05-17
[BT-660] refresh cache upon unlinking

mspector:
Previously, cached access tokens were not cleared when the user account is unlinked. This PR fixes that.

This PR has two commits that test the same behavior:
- one commit defines a new unit test, treating this as an independent behavior
- the other adds functionality to the existing access token caching test

Only one of these commits is needed. Personally, I think it makes sense to test this behavior within the existing tests, but I defer to the reviewers' judgment.
 
---

In all cases:

### DIFF
--- a/tests/unit/bond_test.py
+++ b/tests/unit/bond_test.py
@@ -319,8 +319,11 @@ class BondTestCase(unittest.TestCase):
         access_token, expires_at = bond.get_access_token(self.user_id)
         self.assertEqual(access_token_initial, access_token)
 
-        # unlink and relink account
+        # unlink account and verify that the access token has been cleared
         bond.unlink_account(self.user_id)
+        self.assertIsNone(cache_api.get(self.user_id, namespace=f"{provider_name}:AccessTokens"))
+
+        # relink account
         self.refresh_token_store.save(
             user_id=self.user_id,
             refresh_token_str=refresh_token_renewed,
--- a/tests/unit/bond_test.py
+++ b/tests/unit/bond_test.py
@@ -116,60 +116,96 @@ class BondTestCase(unittest.TestCase):
         self.assertEqual(self.fake_access_token, access_token)
         self.assertEqual(datetime.fromtimestamp(self.expires_at_epoch), expires_at)
 
-    def test_get_access_token_from_cache(self):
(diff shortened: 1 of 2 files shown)


jgainerdewar (approved):
I tend to lean toward more smaller tests that each test narrower pieces of functionality, but I'll defer to Greg's opinion.

mspector:
There's a lot of setup code involved in these tests because the existing setup function isn't well-suited for tests that involve cached access tokens.

The two independent tests are as narrow as I could make it (suggestions welcome for making it narrower), but each one has to basically repeat the access token cache setup.

The single integrated test lets us set up only once, but more importantly, it feels to me like it's testing one piece of functionality: the access token cache.

FYI @gpolumbo-broad if you want to know more about what I was thinking

EDIT: come to think of it, another solution could be to modify the existing setup function so that these tests don't have to be so large. But that's a bigger task

dvoet (approved):

andy7i (approved):
I see the dilemma since the tests are very similar. I think I prefer the initial change (keeping the existing cache test and creating a new test to test the cache after the user unlinks/relinks). That way we're testing one behavior per test, which can be helpful when looking at test failures.

tests/unit/bond_test.py
```diff
@@ -248,6 +284,93 @@ def test_revoke_link_exists(self):
...
+            name="refresh_access_token"
+        )
+
+        # demonstrate that `access_token_initial` has been cached
+        access_token, expires_at = bond.get_access_token(self.user_id)
+        self.assertEqual(access_token_initial, access_token)
+
+        # unlink and relink account
+        bond.unlink_account(self.user_id)
+        self.refresh_token_store.save(
```
andy7i: one thought -- should we check that the `access_token_initial` no longer exists in the cache before relinking the account?

gpolumbo-broad (approved):

tests/unit/bond_test.py
mspector: good idea. I'll add that. thanks Andrew!

andy7i (approved):
Thanks for implementing the suggestion. The PR looks great! 👍

PR #186, opened 2022-05-17
[TOAZ-132] Update bond authentication mechanism to support B2C JWTs

rtitle:
Before, the authentication algorithm was: 

1. Cache google tokeninfo by access token
2. Cache Sam userinfo by google_user_info.user_id

However the only place the google tokeninfo was being used was for the Sam cache. Now we no longer retrieve google tokeninfo, and just cache Sam userinfo by access token. It also checks user enablement in Sam.

This allows us to handle B2C JWTs and Google tokens.

Updated `authentication_test.py` unit tests as part of this. Will also test real tokens via automation tests and manual UI testing with B2C enabled.

--- 

In all cases:


tests/unit/authentication_test.py
```diff
@@ -6,7 +6,7 @@
 from bond_app.authentication import Authentication, AuthenticationConfig, UserInfo
 from bond_app.sam_api import SamApi, SamKeys
 from tests.unit.fake_cache_api import FakeCacheApi
-import json
+import time
 
 
 class TestRequestState:
```
rtitle, line 12: Refactored this test quite a bit based on new logic. Added comments to each test case to hopefully explain what's going on.

rtitle, line 12: Refactored this test quite a bit based on new logic. Added comments to each test case.

rtitle:
fiab-automation-fancy-snake is out of disk space.

jenkins retest

rtitle:
jenkins retest

bond_app/authentication.py
```diff
@@ -76,66 +53,23 @@ def auth_user(self, request_state, token_info_fn=_token_info):
...
-            self.cache_api.add(key='access_token:' + token, value=google_user_info, expires_in=expires_in)
-        else:
-            logging.debug("auth token cache hit for token %s", token)
-
-        sam_user_info = self.cache_api.get(namespace="SamUserInfo", key=google_user_info.id)
+        sam_user_info = None
+
+        # Note: Datastore cache keys have a hardcoded limit of 1500 bytes. JWTs can exceed that, which
+        # unfortunately means we can't cache them.
+        should_cache_token = 1 <= len(token.encode("utf-8")) <= 1500
```
pshapiro4broad: I don't know anything about this codebase, but in general I'd expect a check like this to be done inside the API (`self.cache_api`) instead of in the caller.

rtitle: Thanks, you're right -- will update. I'm amazed you saw this PR. :)

config.ini.ctmpl
```diff
@@ -22,6 +22,20 @@ OPEN_ID_CONFIG_URL=https://storage.googleapis.com/wb-dev-mock-provider/well-know
 USER_NAME_PATH_EXPR=/username
 FENCE_BASE_URL=https://us-central1-broad-dsde-dev.cloudfunctions.net
 
+[anvil]
```
rtitle, line 25: These providers were missing from fiab configs.

bond_app/datastore_cache_api.py
```diff
@@ -47,7 +57,11 @@ def delete_expired_entries():
 
     @staticmethod
     def _build_cache_key(key, namespace):
-        """Create an ndb Key for the key and namespace."""
+        """
+        Create an ndb Key for the key and namespace.
+        Raises InvalidArgument if the cache key is invalid. Note Datastore string cache keys
+        must be at most 1500 bytes.
```
rtitle, line 63: Unfortunately JWTs can exceed 1500 bytes which means they will not be cacheable in Datastore.

bond_app/sam_api.py
```diff
@@ -17,12 +17,20 @@ def user_info(self, access_token):
         """
         headers = {'Authorization': 'Bearer ' + access_token}
         result = requests.get(url=self.base_url + '/register/user/v2/self/info', headers=headers)
+        logging.info('Sam userInfo: status code {}, body {}'.format(result.status_code, result.content))
```
rtitle, line 20: This method now throws 200, 401, or 500 as appropriate.

rtitle:
With @gpolumbo-broad's help I was able to test this in a fiab/local UI. All 4 "mock" providers (fence, dcf-fence, anvil, kids-first) work for all 3 cases:
1. Google auth
2. Google through B2C
3. Azure through B2C

@gpolumbo-broad @dvoet requesting reviews since we've talked about this change. @pshapiro4broad also feel free to review -- even if you're not familiar with the code base it's good to have 👀 to check that things make sense. :)

bond_app/datastore_cache_api.py
dvoet: if we have to, there should be a jti claim in the jwt which is supposed to be unique and could be used as a cache key. But that means verifying the jwt can be trusted first.

dvoet (approved):

bond_app/datastore_cache_api.py
rtitle: Yeah -- I have code to validate JWTs in python [here](https://github.com/DataBiosphere/bond/pull/184) but maybe that can be follow-on work, only if needed.

gpolumbo-broad (approved):

PR #191, opened 2022-09-21
ID-263, ID-263: Enable sts and cache control headers.

Ghost-in-a-Jar:
This pr addresses both https://broadworkbench.atlassian.net/browse/ID-263 and https://broadworkbench.atlassian.net/browse/ID-262

In all cases:

### DIFF
--- a/main.py
+++ b/main.py
@@ -93,3 +93,16 @@ def add_nosniff_content_type_header(response):
     response.headers["X-Content-Type-Options"] = "nosniff"
     response.headers["X-Frame-Options"] = "deny"
     return response
+
+@app.after_request
+def add_strict_transport_security_header(response):
+    # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
+    # recommended settings by mozilla, max-age is 2 years in seconds
+    response.headers["Strict-Transport-Security"] = "max-age=63072000; includeSubDomains; preload"
+    return response
+
+@app.after_request
+def add_cache_control_header(response):
+    # https://grayduck.mn/2021/09/13/cache-control-recommendations/
+    response.headers["Cache-Control"] = "max-age=0, must-revalidate, no-cache, no-store, private"
+    return response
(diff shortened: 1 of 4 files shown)


tlangs (approved):

Shakespeared (approved):

main.py
```diff
@@ -93,3 +93,16 @@ def add_nosniff_content_type_header(response):
     response.headers["X-Content-Type-Options"] = "nosniff"
     response.headers["X-Frame-Options"] = "deny"
     return response
+
+@app.after_request
```
andy7i, line 97: (asking not knowing a lot about flask) -- it seems weird to add a header after the response is sent. Or does `after_request` actually run before the response is sent?

andy7i: found the answer. it's run after the request is processed but before the response is sent...
flask request response timing
https://instructobit.com/tutorial/111/Python-Flask%3A-running-code-before-and-after-every-request

main.py
```diff
@@ -93,3 +93,16 @@ def add_nosniff_content_type_header(response):
...
+def add_strict_transport_security_header(response):
+    # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
+    # recommended settings by mozilla, max-age is 2 years in seconds
```
andy7i, line 100: I took a look through the link (which links to hstspreload.org) and there's something about:
> Once you're confident that there will be no more issues, increase the max-age to 2 years and submit your site to the preload list

from https://hstspreload.org/

I'm not very familiar with the preload list but is it something we should be concerned about?

andy7i (commented):
looks good overall! thanks for adding in the helpful comments too. 

I have a small concern about the hsts preload stuff. If that's fine I can give a thumb. 

Also, I think we can add ID-262 to the title (since this PR addresses 2 tickets)

main.py
Ghost-in-a-Jar: @andy7i I think this is more of a devops thing where they would register the terra ui dns record with that hstspreload list

Ghost-in-a-Jar: it seems to be out of scope of this ticket and pentest requirements

andy7i: one of my concerns was that 2 years might not be the right amount of time to put here. Mozilla says:

> **If** a max-age of 1 year is acceptable for a domain, however, two years is the recommended value as explained on https://hstspreload.org/.

emphasis on `if`. Do we know if 1 year is acceptable?

if you look at https://hstspreload.org/#deployment-recommendations the deployment recommendations slowly ramp up the max-age (starting with 5 minutes and 1 week intervals) .

PR #192, opened 2022-09-21
ID-261 CSRF Mitigation

tlangs:
Remediates https://broadworkbench.atlassian.net/browse/ID-261

Bond needs to implement CSRF protection. To do this, we will provide a random `nonce` in the `state` parameter of the url returned from the `authorization-url` endpoint in Bond. When a client uses that url and gets back an OAuth code, the client must provide that same `nonce` in the state in order for Bond to successfully respond to the `oauthcode` request. 

This is a breaking change, meaning Terra-UI will need to be updated. This cannot merge until Terra-UI forwards state along to bond as a part of the `oauthcode` call.

In all cases:

### DIFF
--- a/bond_app/oauth2_state_store.py
+++ b/bond_app/oauth2_state_store.py
@@ -36,13 +36,15 @@ class OAuth2StateStore:
         ouath2_nonce = OAuth2State(key=OAuth2StateStore._oauth2_state_store_key(user_id, provider), nonce=nonce)
         ouath2_nonce.put()
 
-    def delete_if_exists(self, user_id, provider_name) -> bool:
+    def validate_and_delete(self, user_id, provider_name, nonce) -> bool:
         key = OAuth2StateStore._oauth2_state_store_key(user_id, provider_name)
-        if key.get():
+        oauth2_state = key.get()
+        is_valid = False
+        if oauth2_state:
+            if oauth2_state.nonce == nonce:
+                is_valid = True
             key.delete()
-            return True
-        else:
-            return False
+        return is_valid
 
     # def build_oauth2_state(self, user_id, provider_name):
 
--- a/bond_app/oauth2_state_store.py
+++ b/bond_app/oauth2_state_store.py
@@ -34,8 +34,8 @@ class OAuth2StateStore:
         :param provider:
         :param nonce: random value for csrf protection
         """
-        ouath2_nonce = OAuth2State(key=OAuth2StateStore._oauth2_state_store_key(user_id, provider), nonce=nonce)
-        ouath2_nonce.put()
+        oauth2_nonce = OAuth2State(key=OAuth2StateStore._oauth2_state_store_key(user_id, provider), nonce=nonce)
+        oauth2_nonce.put()
 
     def validate_and_delete(self, user_id, provider_name, nonce) -> bool:
         key = OAuth2StateStore._oauth2_state_store_key(user_id, provider_name)
--- a/bond_app/oauth2_state_store.py
+++ b/bond_app/oauth2_state_store.py
@@ -54,7 +54,9 @@ class OAuth2StateStore:
             decoded_state = json.loads(decoded_raw)
         nonce = secrets.token_urlsafe()
         decoded_state_with_nonce = {**decoded_state, 'nonce': nonce}
-        return base64.b64encode(json.dumps(decoded_state_with_nonce).encode('utf-8')), nonce
+        dumped_state = json.dumps(decoded_state_with_nonce)
+        utf8_dumped_state = dumped_state.encode('utf-8')
+        return base64.b64encode(utf8_dumped_state), nonce
 
     @staticmethod
     def _oauth2_state_store_key(user_id, provider_name):
--- a/bond_app/oauth2_state_store.py
+++ b/bond_app/oauth2_state_store.py
@@ -50,7 +50,8 @@ class OAuth2StateStore:
         if not state:
             decoded_state = {}
         else:
-            decoded_state = json.loads(base64.b64decode(state))
+            decoded_raw = base64.b64decode(state)
+            decoded_state = json.loads(decoded_raw)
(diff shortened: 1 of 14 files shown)


bond_app/oauth2_state_store.py
```diff
@@ -0,0 +1,63 @@
+import base64
+import json
+import secrets
+from collections import namedtuple
+from google.cloud import ndb
+
+# Information associated with a tokens for refreshing service account credentials.
+OAuth2StateInfo = namedtuple('OAuth2StateInfo', ['nonce'])
```
Ghost-in-a-Jar: There's finally a better way to do this type of thing:
https://realpython.com/python-data-classes/
ex:
```
from dataclasses import dataclass
@dataclass
class OAuth2StateInfo:
    nonce: str
    
```

bond_app/oauth2_state_store.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def save(self, user_id, provider, nonce):
+        """
+        Persists a RefreshToken by creating a new entity or updating an existing entity with the same id
+        :param user_id
+        :param provider:
+        :param nonce: random value for csrf protection
+        """
+        ouath2_nonce = OAuth2State(key=OAuth2StateStore._oauth2_state_store_key(user_id, provider), nonce=nonce)
+        ouath2_nonce.put()
```
Ghost-in-a-Jar: what does this do?

Ghost-in-a-Jar: oh also typo heh

Ghost-in-a-Jar: oh i see this persists it in the datastore, what is the underlying db we are using here?

tlangs: There is no underlying _real_ DB. We're using https://cloud.google.com/datastore/

Ghost-in-a-Jar: ok gotcha 👍

bond_app/oauth2_state_store.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def validate_and_delete(self, user_id, provider_name, nonce) -> bool:
+        key = OAuth2StateStore._oauth2_state_store_key(user_id, provider_name)
+        oauth2_state = key.get()
+        is_valid = False
+        if oauth2_state:
+            if oauth2_state.nonce == nonce:
+                is_valid = True
+            key.delete()
+        return is_valid
```
Ghost-in-a-Jar: nitpicky: Id say something like this is slightly more readable:
```
oauth2_state = key.get()
if oauth2_state is None:
    return False
else:
    key.delete()
    return oauth2_state.nonce == nonce
```

bond_app/oauth2_state_store.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def state_with_nonce(self, state):
+        if not state:
+            decoded_state = {}
+        else:
+            decoded_state = json.loads(base64.b64decode(state))
+        nonce = secrets.token_urlsafe()
+        decoded_state_with_nonce = {**decoded_state, 'nonce': nonce}
+        return base64.b64encode(json.dumps(decoded_state_with_nonce).encode('utf-8')), nonce
```
Ghost-in-a-Jar: Just for the sake of making stack traces easier to read id say split the json dumps and encoding on different lines

tests/unit/fake_oauth2_state_store.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def state_with_nonce(self, state):
+        if not state:
+            decoded_state = {}
+        else:
+            decoded_state = json.loads(base64.b64decode(state))
+        nonce = self.test_nonce
+        decoded_state_with_nonce = {**decoded_state, 'nonce': nonce}
+        return base64.b64encode(json.dumps(decoded_state_with_nonce).encode('utf-8')), nonce
+
```
Ghost-in-a-Jar, line 53: same 2 comments as in the actual class

Ghost-in-a-Jar (approved):
I think this is good and esp if you have tested it manually locally -- one nice to have though would be a last test (and more of these in the repo in general) that used a flask test client to actually stand up the app and call endpoints. https://flask.palletsprojects.com/en/1.1.x/testing/#the-first-test

tlangs:
I think @gpolumbo-broad might have words to say about adding more "Integration" or "end-to-end" tests to the repo. 😅

bond_app/oauth2_state_store.py
tlangs: oh cool! Like Records! Or Case Classes!

tlangs:
@Ghost-in-a-Jar Turns out we actually have "automation" tests that test a deployed branch of Bond upon PR.

bond_app/oauth2_state_store.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class OAuth2StateStore:
+    """
+    Stores OAuth2 State nonces for csrf protection.
+    """
+
+    def save(self, user_id, provider, nonce):
+        """
+        Persists a RefreshToken by creating a new entity or updating an existing entity with the same id
```
gpolumbo-broad: RefreshToken?

swagger/api-docs.yaml
```diff
@@ -74,7 +74,6 @@ paths:
       summary: Builds an OAuth authorization URL that a user must use to initiate the OAuth dance.
```
andy7i:
```suggestion
      summary: Builds an OAuth authorization URL that a user must use to initiate the OAuth dance. Creates a 1 time use link. There can only be 1 active link per user per provider. The last link created will invalidate any prior links.
```

andy7i (commented):
partial review since Greg is almost done reviewing

tlangs:
jenkins retest

gpolumbo-broad (approved):

PR #213, opened 2023-08-22
[ID-760] Use blessed base image

em-may:
See https://broadinstitute.slack.com/archives/C03GMG4DUSE/p1692732291046169

### DIFF
--- a/bond_app/jwt_token.py
+++ b/bond_app/jwt_token.py
@@ -12,7 +12,7 @@ class JwtToken:
         :param encoded_str: encoded JWT
         :param user_name_path_expr: forward slash delimited path to user name within JWT
         """
-        self.raw_dict = jwt.decode(b64decode(encoded_str), algorithms=['HS256'], options={'verify_signature': False})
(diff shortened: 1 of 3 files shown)


bond_app/jwt_token.py
```diff
@@ -11,7 +11,7 @@ def __init__(self, encoded_str, user_name_path_expr):
         :param encoded_str: encoded JWT
         :param user_name_path_expr: forward slash delimited path to user name within JWT
         """
-        self.raw_dict = jwt.decode(encoded_str, verify=False)
+        self.raw_dict = jwt.decode(encoded_str, algorithms=['HS256'], options={'verify_signature': False})
```
em-may, line 14: These lines are equivalent, PyJWT>= 2.0 uses an options dict with a key called `verify_signature` to replace the arg called `verify`. See the api docs [here](https://pyjwt.readthedocs.io/en/stable/api.html) for more details.

Ghost-in-a-Jar (approved):

PR #220, opened 2023-08-31
ID-754 Update google dependencies

Ghost-in-a-Jar:
The google dependencies are way behind.
---

In all cases:

- [ ] **Release to production** - Releasing to prod is a manual process (see [instructions to release to prod](https://github.com/DataBiosphere/bond/blob/develop/README.md#production-deployment-checklist)). Either do this now or create a ticket and schedule the release.


em-may:
We _can_ do this at a later date if we want because it does seem to work, but it should definitely be after Bond in prod is successfully deployed to k8s. It may be better to rewrite Bond in Java instead.

em-may (approved):

PR #229, opened 2024-01-10
ID-1011 Make Bond only return Status 500 if critical subsystems are down

tlangs:
What:

https://broadworkbench.atlassian.net/browse/ID-1011

If a fence provider goes down, don't take Bond down with it.

Why:

Bond fails for everyone if one provider goes down.

How:

Make Bond only restart if Sam or Datastore goes down.

---

In all cases:

- [ ] **Release to production** - Releasing to prod is a manual process (see [instructions to release to prod](https://github.com/DataBiosphere/bond/blob/develop/README.md#production-deployment-checklist)). Either do this now or create a ticket and schedule the release.

### DIFF
--- a/bond_app/routes.py
+++ b/bond_app/routes.py
@@ -20,7 +20,7 @@ from . import fence_token_storage
 from .open_id_config import OpenIdConfig
 from .sam_api import SamApi
 from .oauth_adapter import OauthAdapter
-from .status import Status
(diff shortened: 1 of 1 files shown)


em-may (approved):
We should make proper probe endpoints in the future but Bond in prod is broken now soooo

Shakespeared (approved):

